PR #5, opened 2026-03-26
feat: QJL ghost detection, distortion bounds, 7-signal quality

MaTriXy:
## Summary

- **QJL ghost token detector** — 1-bit sketch clustering finds wasteful near-duplicate runs (~40% better sensitivity)
- **Distortion bounds metric** — theoretical quality ceiling based on TurboQuant rate-distortion theory
- **7-signal quality scoring** — 2 new signals (Message Efficiency + Compression Opportunity)

## Files changed

- `openclaw/src/jl-sketcher.ts` — QJL 1-bit sketch library (new)
- `openclaw/src/waste-detectors.ts` — GhostTokenQJL detector (#8)
- `openclaw/src/quality.ts` — distortion bounds, 2 new signals, weight rebalancing
- `docs/turboquant-enhancements.md` — documentation (new)

## Test plan

- [ ] Verify ghost detection on sample session data
- [ ] Validate distortion bounds across context window sizes
- [ ] Check quality score weights sum to 100%

### DIFF
--- a/docs/turboquant-enhancements.md
+++ b/docs/turboquant-enhancements.md
@@ -7,7 +7,7 @@ Three new features inspired by [TurboQuant](https://research.google/blog/turboqu
 | Enhancement | File | What It Does |
 |-------------|------|-------------|
 | QJL Ghost Token Detector | `openclaw/src/jl-sketcher.ts` (new) + `waste-detectors.ts` | Sketch-based clustering to find wasteful near-duplicate runs |
-| Distortion Bounds Metric | `openclaw/src/quality.ts` | Theoretical quality ceiling based on TurboQuant distortion theory |
+| Distortion Bounds Metric | `openclaw/src/quality.ts` | Estimated quality ceiling inspired by TurboQuant distortion concepts |
 | Two-Stage Quality Scoring | `openclaw/src/quality.ts` | 2 new signals: Message Efficiency + Compression Opportunity |
 
 ---
@@ -90,11 +90,11 @@ For each cluster with 2+ runs:
 
 ### Problem
 
-Users optimize their quality score without knowing the theoretical ceiling. If you're at 85/100 and the theoretical max for your configuration is 88, further optimization is futile — structural changes (shorter sessions, different model) are needed.
+Users optimize their quality score without knowing the estimated ceiling. If you're at 85/100 and the estimated max for your configuration is 88, further optimization is unlikely to help — structural changes (shorter sessions, different model) are needed.
 
 ### Solution
 
-`computeDistortionBounds()` calculates a theoretical quality ceiling based on TurboQuant's distortion-rate theory, adapted to context windows.
+`computeDistortionBounds()` calculates an estimated quality ceiling using a heuristic inspired by TurboQuant's distortion-rate concepts, adapted to context windows. This is a useful approximation, not a proven mathematical limit.
 
 ### API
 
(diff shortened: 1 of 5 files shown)


alexgreensh (changes requested):
Hey @MaTriXy, first off, welcome and thank you for this contribution! The detection concepts here are solid, especially the ghost token clustering and the message efficiency signal. I'm actually planning to adapt some of these ideas for the Claude Code (Python) side of the project too.

A few issues need fixing before we can merge. Grouped by priority:

### 🔴 Build-breaking (must fix)

**1. `scoreToGrade()` removal breaks cli.ts and dashboard.ts**

Removing `scoreToGrade()` and the `grade` field from `QualityReport` and `scoreSessionQuality()` will break at least 7 call sites:
- `cli.ts:260` displays `report.grade`
- `dashboard.ts:11` imports `scoreToGrade`
- `dashboard.ts:225` reads `sq.grade`
- `dashboard.ts:341` calls `scoreToGrade(score)`
- Plus multiple render points in the dashboard HTML

**Fix:** Please keep `scoreToGrade()` and the `grade` field on both interfaces. If you'd like to deprecate grades in favor of bands, we can do that in a separate PR with a migration.

**2. Tier mismatch**

The docs say the ghost detector is Tier 2, but `ALL_DETECTORS` registers it as Tier 3. Please align them (I'd suggest Tier 2 to match the other session-analysis detectors).

### 🟡 Silent bugs (should fix)

**3. Missing recommendation cases for new signals**

`generateQualityRecommendations()` uses a `switch` on signal names but has no cases for "Message Efficiency" or "Compression Opportunity". When these score below 70, users get no guidance. You've already written the recommendation text in the signal descriptions, so just add matching cases to the switch.

**4. Double computation in `computeDistortionBounds()`**

`computeDistortionBounds()` internally computes all 7 signals to get `achievedScore`, but `scoreQuality()` already computes them and then *overwrites* that value. The signals run twice and the first result gets thrown away.

**Fix:** Accept pre-computed signals as an optional parameter:

```typescript
export function computeDistortionBounds(
  runs: AgentRun[],
  modelContextWindow: number,
  precomputedSignals?: QualitySignal[]
): DistortionBounds
```

**5. Add weight-sum validation**

With 7 signals now, a runtime check that weights sum to 1.0 would prevent future drift:

```typescript
const sum = signals.reduce((s, sig) => s + sig.weight, 0);
if (Math.abs(sum - 1.0) > 0.001) throw new Error(`Weights sum to ${sum}`);
```

### 💭 Suggestions (non-blocking, worth discussing)

**6. Distortion bounds framing**

The `1/sqrt(effectiveCapacity)` formula is a reasonable heuristic, but calling it a "theoretical ceiling" based on "TurboQuant distortion theory" oversells it. Could you soften the framing to "estimated quality ceiling" or "heuristic upper bound"? Users should understand it's a useful approximation, not a proven mathematical limit.

**7. Sketch complexity vs. simple grouping**

For the ghost token detector, have you considered simple field-based grouping on `(agentName, model, runType)` instead of sketch clustering? The existing `AgentRun` metadata already gives you deterministic grouping without the hash/similarity machinery. The sketcher is well-written, but it's a lot of algorithmic surface area for a problem that might be solvable with a simple Map. Happy to discuss if you see cases where field grouping would miss things that sketches catch.

**8. `sketchSimilarity` length guard**

If two sketches with different dimensions get compared, the result is silently wrong. Worth adding a length check that throws on mismatch.

---

Excited to see v2! The ghost token clustering concept fills a real gap in our detection suite, and with the fixes above this should merge cleanly.

alexgreensh (changes requested):
Hey @MaTriXy, really impressed by the turnaround here. You addressed every point thoroughly, and fair point on items 3 and 8, both were already handled in your original code. The dual-strategy approach for the ghost detector is actually better than what I suggested, clean separation with simple grouping as the sensible default.

One change needed before I can merge:

**Please revert the `measure.py` commit (7970992).**

I appreciate the initiative! But the Python scorer operates on a fundamentally different data model (JSONL session tuples vs structured `AgentRun` objects), so the signals need a different implementation approach. The weight rebalancing would also silently change scores for all existing Claude Code plugin users. I'd like to handle the Python adaptation separately so I can design it against the JSONL data we actually have.

The OpenClaw TypeScript changes all look good to me. Once the measure.py commit is reverted, I'll merge.

Thanks again for a really solid contribution.

PR #12, opened 2026-04-05
fix: recover from /statusline clobber with SessionStart hint (v3.5.1)

alexgreensh:
## What

When a user runs Claude Code's built-in `/statusline` slash command, it rewrites the `statusLine` key in `~/.claude/settings.json`, silently overwriting Token Optimizer's quality bar. The cache hook keeps running (quality data flows to disk), but nothing displays it. No warning, no clue where to look.

A user just hit this and asked how to recover. This patch makes the recovery self-service.

## How

Extends the `ensure-health` SessionStart branch to distinguish three statusline states instead of two:

| State | Before | After |
|---|---|---|
| No `statusLine` at all | Install ours silently | Install ours silently (unchanged) |
| Our statusline + missing cache hook | Reinstall | Reinstall (unchanged) |
| **Foreign statusline + our cache hook still present** | **Silent, no recovery path** | **Prints: `Restore with: measure.py setup-quality-bar`** |

Never overwrites a user's chosen statusline. The hint is idempotent and honors the existing `quality_bar_disabled` config for permanent opt-out.

## Why not auto-reinstall?

Because the user may have replaced our statusline on purpose. We detect but don't overwrite. The fix is one command away and the user decides.

## Files

- `skills/token-optimizer/scripts/measure.py` — defensive branch in the existing `ensure-health` block (line ~10554)
- `README.md` — troubleshooting subsection under **Live Quality Bar** + version badge bump
- `.claude-plugin/plugin.json`, `.claude-plugin/marketplace.json` — version bump 3.5.0 → 3.5.1

## Verification

Ran a local 4-scenario harness against the real `measure.py ensure-health` path with a sandboxed `HOME`:

```
Scenario 1 (clobber):  rc=0 hint_printed=True   ✓
Scenario 2 (intact):   rc=0 hint_printed=False  ✓
Scenario 3 (empty):    rc=0 hint_printed=False  ✓
Scenario 4 (no hook):  rc=0 hint_printed=False  ✓
```

## Background context

This is the second time the `statusLine` key has been nuked from outside Token Optimizer. First was the GSD plugin removal incident in March. `/statusline` is the new foot-gun. README troubleshooting entry covers both.

## Checklist

- [x] Version bumped in plugin.json + marketplace.json
- [x] README badge bumped
- [x] Local scenarios verified (4/4 pass)
- [x] No tests added to public repo
- [x] No Co-Authored-By trailer (CLA-gated repo)
- [x] Defensive branch only — no behavioral change for users whose statusline is still Token Optimizer's

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -13,7 +13,7 @@
       "name": "token-optimizer",
       "source": "./",
       "description": "Audit, fix, and monitor Claude Code context window usage. Find the ghost tokens.",
-      "version": "3.5.0",
+      "version": "3.5.1",
       "author": {
         "name": "Alex Greenshpun",
         "url": "https://linkedin.com/in/alexgreensh"
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -7,7 +7,7 @@
   },
   "homepage": "https://github.com/alexgreensh/token-optimizer",
   "repository": "https://github.com/alexgreensh/token-optimizer",
-  "version": "3.5.0",
+  "version": "3.5.1",
   "license": "AGPL-3.0-only",
   "keywords": ["token", "optimization", "context", "audit", "cost", "coach"]
 }
--- a/README.md
+++ b/README.md
@@ -328,15 +328,19 @@ A glance at your terminal tells you if you're in trouble. Colors shift from gree
 python3 measure.py setup-quality-bar      # one-time install
 ```
 
-**My quality bar disappeared — how do I get it back?** Running Claude Code's built-in `/statusline` rewrites the `statusLine` key in `~/.claude/settings.json` and can overwrite Token Optimizer's entry. Since v3.5.1, SessionStart detects this and prints a one-line recovery hint. Restore it with:
+**My quality bar disappeared — how do I get it back?** Running Claude Code's built-in `/statusline` rewrites the `statusLine` key in `~/.claude/settings.json` and silently overwrites Token Optimizer's entry. Since v3.5.1, SessionStart detects this and **auto-restores** the quality bar — just start a new session and it's back. You'll see a one-line notice explaining what happened.
+
+**I really don't want the quality bar anymore — how do I turn it off for good?** Run:
 
 ```bash
-python3 measure.py setup-quality-bar --status     # check what's missing
-python3 measure.py setup-quality-bar --dry-run    # preview the diff
-python3 measure.py setup-quality-bar              # apply
+python3 measure.py setup-quality-bar --uninstall
 ```
 
-If you replaced the statusline on purpose, the recovery hint is idempotent and harmless — ignore it, or silence it permanently by setting `"quality_bar_disabled": true` in `~/.claude/token-optimizer/config.json`.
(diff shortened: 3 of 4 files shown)


alexgreensh:
Follow-up per Alex's call: flip from print-hint to **auto-restore**, and make `--uninstall` **sticky**.

## What changed in `d441706`

1. **ensure-health auto-restores** the quality bar when the clobber condition hits (foreign `statusLine` + our cache hook still present). Prints a one-line notice so the user knows what happened, then they just keep working — statusline is back without them running anything.

2. **`setup_quality_bar()` gains a `force` parameter.** Default `False`, so direct CLI installs still respect a pre-existing foreign `statusLine` (unchanged behavior). Only `ensure-health`'s recovery path calls it with `force=True`, and only after confirming the cache hook survived — which is strong evidence the user had our quality bar installed and something overwrote just the display layer.

3. **`--uninstall` is now sticky.** New helper `_set_quality_bar_disabled()` writes `quality_bar_disabled: true` to `~/.claude/token-optimizer/config.json`. Both `ensure-health` and the `quality-cache` self-heal already gated on this flag — they just had no writer until now. So the opt-out survives across sessions.

4. **Explicit install clears the flag.** Running `setup-quality-bar` with no args sets `quality_bar_disabled: false`, so "I changed my mind, bring it back" works without hand-editing config.

5. **Natural-language opt-out path documented** in README: user says "remove the Token Optimizer statusline" to Claude Code → Claude runs `measure.py setup-quality-bar --uninstall` → flag sticks → done.

## How we know clobber vs intent

| Signal | Interpretation |
|---|---|
| `quality_bar_disabled: true` in `config.json` | User explicitly opted out. Respect forever. Never auto-restore. |
| Flag absent/false, statusLine is ours, cache hook installed | Normal operation. Do nothing. |
| Flag absent/false, statusLine missing, cache hook missing | First-run or fresh setup. Install silently. |
| Flag absent/false, statusLine missing, cache hook present | Partial clobber. Reinstall statusline. |
| **Flag absent/false, foreign statusLine, cache hook still present** | **Clobber (e.g., `/statusline`). Auto-restore with force.** |
| Flag absent/false, foreign statusLine, cache hook absent | User truly has their own setup. Do nothing (direct install shows integration instructions). |

## Verification

Extended local scenario harness from 4 to 7, all pass:

```
1 clobber->restore:              restored=True  notice=True   OK
2 opted-out -> no-op:             restored=False               OK
3 uninstall:                      sl_removed=True flag_set=True OK
4 sticky after ensure-health:     sl_absent=True               OK
5 reinstall clears flag:          flag_cleared=True sl_back=True OK
6 direct install respects custom: preserved=True               OK
7 intact setup stays quiet:       notice_absent=True           OK
```

## Updated diff stat for the whole PR

```
 .claude-plugin/marketplace.json           |   2 +-
 .claude-plugin/plugin.json                |   2 +-
 README.md                                 |  24 ++++++-
 skills/token-optimizer/scripts/measure.py |  83 +++++++++++++++++++++---
 4 files changed, 92 insertions(+), 19 deletions(-)
```

Still well under a day's work to review, still no tests in the public repo, still no `Co-Authored-By` trailer.

PR #35, opened 2026-05-27
security: harden hook paths, glob boundaries, credential redaction, and session isolation

GalitGal:
## Summary

Security hardening across 7 findings identified during independent review. All changes are fail-open (hook errors never block the user's tool call) and backward-compatible.

### H1 — bash_hook.py: Path derivation cross-check
When `CLAUDE_PLUGIN_ROOT` is set by the dispatcher, verify that the `__file__`-derived `compress_path` and `launcher_path` land within the declared plugin root. A mismatch (symlink swap or relocated install) triggers fail-closed exit. Primary path derivation stays `__file__`-based — not env-var-based — to avoid a different injection vector.

### H2 — structure_replay.py: Post-glob boundary filter
`_gather_input_paths()` now validates every `glob.glob()` expansion against a safe root (`~/.claude` by default, overridable via `TOKEN_OPTIMIZER_SAFE_ROOT`). Expansions that escape the root are silently dropped. Prevents crafted glob patterns from reaching `~/.ssh/`, `/etc/`, etc.

### M1 — python-launcher.sh: Interpreter allowlist
Adds `_is_safe_prefix()` validation before accepting any interpreter binary from PATH search. Accepted prefixes: `/usr/bin`, `/usr/local/bin`, `/opt/homebrew/bin`, `/opt/homebrew/opt`, and the dynamic `brew --prefix` (gracefully skipped on Linux). Prevents PATH-order attacks where a malicious directory appears before a legitimate one.

### M2 — context_intel.py: Sensitive path filter
`_extract_paths()` now calls `_is_sensitive_path()` before logging any path to SQLite. Blocked segments: `/.ssh/`, `/.aws/`, `/.gnupg/`, `/etc/`, `/run/secrets`, `credentials`, `secrets`, `.env`, `id_rsa`, `id_ed25519`, `.pem`, `.key`.

### M3-a — archive_result.py: Credential redaction before write
Imports `_TOKEN_PATTERNS` from `bash_compress` (same 22-pattern list that guards compression output) and applies `_redact_credentials()` on the tool response before every `json.dump`. No credential plaintext ever reaches disk — redaction happens before the `os.open` call.

### M3-b — archive_result.py: Archive TTL enforcement
Adds `cleanup_old_archives(max_age_hours=48)` function, called at the top of `archive_result()` on every PostToolUse invocation. Best-effort (errors swallowed). Matches the 48h TTL already applied to session-store DBs.

### M4 — session_store.py: UUID fallback for invalid session IDs
Replaces the static `return "unknown"` fallback with `return f"fallback-{uuid.uuid4().hex[:12]}"`. Prevents all sessions with missing or malformed IDs from sharing one SQLite database, which would allow cross-session data leakage.

### L2 — run.py: proc.poll() guard before proc.kill()
Adds `if proc.poll() is None:` before `proc.kill()` in the TimeoutExpired handler. Avoids sending SIGKILL to a process that exited naturally between the timeout and the kill call, preventing a theoretical reused-PID scenario on some POSIX implementations.

## Deferred (L1)
Out-of-band checksum integrity for `install.sh` is tracked separately — it requires release infrastructure (signed releases, external hash hosting) not present in this repo. Filed as an issue on the fork.

## Sprint contract
All 11 machine-verifiable criteria (C1–C11) pass by grep inspection.

### DIFF
--- a/hooks/python-launcher.sh
+++ b/hooks/python-launcher.sh
@@ -10,6 +10,35 @@
 
 set -eu
 
+# Known-safe prefixes for Python interpreter binaries.
+# Binaries outside these directories are rejected even if on PATH.
+# This prevents a compromised PATH entry from hijacking the interpreter.
+_SAFE_PREFIXES="/usr/bin /usr/local/bin /opt/homebrew/bin /opt/homebrew/opt"
+
+# Populate brew prefix once (non-fatal on Linux where brew is absent).
+_BREW_PREFIX=""
+if command -v brew >/dev/null 2>&1; then
+    _BREW_PREFIX="$(brew --prefix 2>/dev/null || true)"
+fi
+
+_is_safe_prefix() {
+    local binpath="$1"
+    # Check static safe prefixes first.
+    local prefix
+    for prefix in $_SAFE_PREFIXES; do
+        case "$binpath" in
+            "$prefix"/*) return 0 ;;
+        esac
+    done
+    # Also allow the brew prefix when available.
+    if [ -n "$_BREW_PREFIX" ]; then
+        case "$binpath" in
+            "$_BREW_PREFIX"/*) return 0 ;;
+        esac
+    fi
+    return 1
+}
+
 find_interpreter() {
     local name="$1"
     local IFS=:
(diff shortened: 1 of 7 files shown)


alexgreensh:
Hi Galit! 👋

Thank you for this thorough security review. I've gone through all 7 changes and they look solid. Clean, well-scoped, fail-open design, backward-compatible.

Quick notes on the review:
- **H1** (path cross-check) and **H2** (glob boundary filter) are excellent catches
- **M1** (interpreter allowlist) is a nice defense-in-depth layer
- **M2-M4** all address real data hygiene concerns
- **L2** (proc.poll guard) is a good defensive fix

The L1 (out-of-band checksums) you deferred has been addressed separately in v5.7.2 (issue #36, just shipped).

To move this forward, could you sign the CLA? Just post a comment with exactly:

Looking forward to merging this!

GalitGal:

alexgreensh:
Quick fix: the CLA check was using exact string matching, so your signature got rejected because of trailing whitespace from the browser. Just pushed a fix. Could you comment `recheck` to retrigger? Should work now. 🙏

PR #37, opened 2026-05-27
fix: reset IFS in _is_safe_prefix to prevent inheritance from find_interpreter

danikdanik:
`find_interpreter` sets local IFS=: to split PATH by colons. In bash, local variables have dynamic scope, so _is_safe_prefix inherits IFS=: when called from `find_interpreter`. 
This means $_SAFE_PREFIXES (space- separated) is treated as one big token instead of five prefixes, and every safe-prefix check silently fails -- rejecting all Python interpreters as unsafe.

Adding local IFS=$' \t\n' at the top of _is_safe_prefix resets it to the bash default before the loop so the prefixes split correctly.

Reproduced on macOS with Python via Homebrew (/opt/homebrew/bin/python3)


danikdanik:

PR #39, opened 2026-05-27
fix: use shutil.rmtree for archive cleanup so nested dirs are reclaimed

danikdanik:
The prior cleanup loop called `entry.unlink()` per child, which raises IsADirectoryError on subdirectories. The error was caught and suppressed, then `session_dir.rmdir()` failed because the directory was non-empty (also caught and suppressed).
Result: any session archive directory that ever contained a subdirectory persisted past the 48-hour TTL forever.

Switching to shutil.rmtree(..., ignore_errors=True) handles nested content, drops the dual-suppress pattern, and keeps the "removed" counter honest by re-checking session_dir.exists() before incrementing.


alexgreensh:
Thanks for catching this @danikdanik! The nested directory cleanup bug was real and affecting users.

We incorporated your fix with additional security hardening in v5.7.9:
- Added symlink guards before rmtree to prevent symlink-swap attacks (the old unlink() loop was actually safe against symlinks, but rmtree follows them)
- Used `os.lstat()` instead of `stat(follow_symlinks=False)` for Python 3.9 compatibility
- Also hardened the same pre-existing pattern in measure.py's `archive_cleanup()`

Closing this PR since we shipped the fix directly, but your bug report and approach were spot on.

PR #43, opened 2026-05-28
fix: redact credentials in MCP replacement output before printing to stdout

danikdanik:
Found this while reviewing the recent `archive_result.py` changes. `_redact_credentials` runs at line 283 and the result (`safe_response`) gets written to the on-disk archive and into the SessionStore preview, which is correct. But a few lines later, the MCP-tool branch hands the *unredacted* `tool_response` to `_detect_output_type` and `_compress_mcp_preview`, and that compressed preview is what we print as `updatedMCPToolOutput`. So the archive on disk is clean, but the live context the assistant sees still has the raw token.

Two-line fix: swap those two arguments to `safe_response`. I left the char/token counters on `tool_response` since they describe the original size, and I left the output hash on `tool_response` too because hashing the redacted form would collide unique payloads that happen to share the same redacted projection.

### Repro

Set up an MCP tool that returns more than 4096 chars and includes something `_TOKEN_PATTERNS` already catches, e.g. an `Authorization: Bearer sk-ant-api03-...` header. Look at the PostToolUse stdout JSON.

Before: the token appears verbatim inside `updatedMCPToolOutput`.
After: the token is `[REDACTED]`, matching what was already on disk.

Sanity-check both ends by grepping `tool-archive/<session>/<id>.json` for `[REDACTED]` (present in both) and the stdout JSON for the token string (present before, gone after).

### What this closes

Everything `_TOKEN_PATTERNS` already catches now flows through the live channel too: AWS access keys, Anthropic `sk-ant-*`, GitHub PATs (classic / OAuth / server / refresh / fine-grained), npm tokens, Stripe live and restricted, HuggingFace, generic Bearer headers, Google API and OAuth, JWTs, PEM private-key headers, DB URIs with embedded passwords, basic-auth URLs.

### What this does NOT close - TODO (@alexgreensh FYI):

Two independent regex bugs in `bash_compress.py`,should be separately fixed:

* `sk-proj-*` OpenAI project-scoped keys. The regex `sk-[a-zA-Z0-9]{20,}` excludes hyphens, so the match fails the length quantifier at the hyphen in `proj-` and nothing gets redacted. 

 > Reference: https://github.com/gitleaks/gitleaks/pull/1780

* Slack 4-segment tokens. The pattern `xoxb-[0-9]+-[a-zA-Z0-9]+` only captures the first two segments; the third high-entropy segment survives unredacted.

> Reference: https://github.com/trufflesecurity/trufflehog/pull/3249


alexgreensh:
Thanks @danikdanik! Merged the credential redaction fix in v5.7.12 with CE review.

Also noted your two regex gap findings (`sk-proj-*` OpenAI keys and Slack 4-segment tokens) — both confirmed real by our security review. We'll fix those separately.

CE security review also found a third gap: the legacy `archive_result()` in measure.py has no redaction step at all (only reachable via direct CLI invocation, not the hook path). Adding that to the fix queue too.

PR #45, opened 2026-05-28
Extract shared path resolver to eliminate duplication across skill files

danikdanik:
## Summary
Consolidates duplicated path-resolution logic across multiple scripts into a single shared resolver module, reducing code duplication and improving maintainability.

## Changes
- Extracts common path-resolution prelude into a dedicated resolver script
- Updates dependent scripts to use the shared resolver
- Maintains backward compatibility with existing functionality

The resolver is 111 lines but it lives in hooks/ and **isn't loaded** into LLM context the way skill bodies are, so the LLM-visible delta is almost -100 .

### DIFF
--- a/commands/health.md
+++ b/commands/health.md
@@ -8,27 +8,13 @@ Run a session health check and help the user manage running sessions safely.
 
 ## Steps
 
-1. Resolve measure.py path:
+1. Resolve measure.py path (shared resolver):
 ```bash
-RUNTIME="${TOKEN_OPTIMIZER_RUNTIME:-}"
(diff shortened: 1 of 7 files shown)


alexgreensh:
Closing: the bootstrap primary path (`~/.claude/token-optimizer/`) points to the data directory on symlink/dev installs, not the repo. resolve.sh wouldn't be found there, breaking all 6 skills simultaneously. The ls glob fallback also picks the oldest cached version (lexicographic sort), not the active one. The inline duplication is annoying but independently resilient — each skill works even when others break. Thanks for the cleanup effort though.

PR #46, opened 2026-05-28
perf: skip Codex adapter imports on the Claude Code path

danikdanik:
## MOtivation
measure.py is imported by every Token Optimizer hook (including the per-prompt UserPromptSubmit hook). Its two Codex adapters (codex_io, codex_session) were imported at module top but are only needed on the Codex runtime, so on the Claude Code path they were loaded for nothing.

## The gain (measured) 
The before/after table (cold ~273→245 ms, warm ~33→31 ms, bare import measure no longer loads codex), the one-line time command, and the honest "win is modest / ~222 ms is measure.py's own compile, deferred" note 
## The change 
Single added import per site, no logic/syntax altere

### DIFF
--- a/skills/token-optimizer/scripts/measure.py
+++ b/skills/token-optimizer/scripts/measure.py
@@ -92,8 +92,11 @@ from hook_io import read_stdin_hook_input as _read_stdin_hook_input_shared
 from plugin_env import resolve_plugin_data_dir
 from runtime_env import claude_home, detect_runtime, runtime_home, runtime_name_for_humans
 
-import codex_io
-import codex_session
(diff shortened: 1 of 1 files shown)


alexgreensh:
Closing: the lazy import approach has a concrete defect in `checkpoint_trigger` (unconditional import outside any conditional, defeats the optimization on the highest-frequency hook path). Beyond that, scattering 12 local imports creates a NameError landmine for future contributors — misses are invisible in Claude Code testing and only surface on Codex. That said, the underlying concern is valid — we're going to investigate the import cost properly and see if there's a cleaner way to speed up the hook path. Thanks for the profiling work.

PR #62, opened 2026-06-15
fix(parser): clamp negative/non-finite token counts at the parse boundary

danikdanik:
## Why

`parseSession` / `parseSessionTurns` coalesce missing usage fields with `?? 0`, which only catches `undefined`/`null`. A corrupt or hand-edited JSONL line carrying a **negative** (or `NaN`/`Infinity`/string) token count passed straight through into the running totals and the cost computation, so a single bad line could poison an entire session's numbers. `savings.ts` already guards its own data boundary with `numClamp()`; the main parse path had no equivalent.

## What

- Add a small `nonNeg()` helper that coerces with `Number()` then clamps to a non-negative finite value.
- Apply it to the resolved token fields in **both** `parseSession` and `parseSessionTurns` (input, output, cache-read, cache-write, 1h/5m cache-write). `parseSessionTurns` feeds `calculateCost` and the `extractCostlyPrompts` ranking, so it carried the identical bug.
- `explicitCacheRead` / `promptCached` are deliberately left possibly-`undefined` so the `inputForCost` / `freshInputTokens` branch can still distinguish "field present" from "field absent".
- `Number()` coercion also recovers a string-typed count (`"1234"`) instead of discarding it.

## How (verified)

A session mixing `input_tokens: -500`, `Infinity`, and a string `"1000"` now yields finite, non-negative totals and per-turn costs in both parse paths (negatives/Infinity → 0, `"1000"` → 1000). `tsc` build is green; committed `dist/` rebuilt.

## Deliberately not addressed

`nonNeg` overlaps `numClamp()` in `savings.ts`; consolidating into a shared module is left out to avoid introducing a `session-parser` ↔ `savings` dependency cycle in this focused change.


alexgreensh:
Merged in v5.11.13 — the negative/non-finite clamp closes a real cost-accuracy hole. Thanks, Dani.

PR #64, opened 2026-06-15
fix(memory): bound in-memory per-session tracking against unclean session ends

danikdanik:
## Why

Three continuity/fresh-nudge collections in `index.ts` (`_continuityInjectedSessions`, `_freshNudgeFiredSessions`, `_freshNudgePriorScores`) and the `states` Map in `checkpoint-policy.ts` are pruned on `session:end`. A session that ends uncleanly (process killed, network disconnect) never fires that event, so its entries leak for the gateway's entire lifetime. On an always-on gateway these grow without bound.

## What

Add oldest-first eviction caps (`MAX_TRACKED_SESSIONS = 2000`):

- `index.ts` gets `boundedSetAdd()` / `boundedMapSet()` helpers. The Map variant refreshes recency on update (`_freshNudgePriorScores` is re-set every patch, so this keeps an actively-scored session from being evicted). The Set guards are add-once fire-guards, so they intentionally do not refresh — a refresh would be dead code.
- `checkpoint-policy.ts` gets `rememberState()`.
- The `session:end` cleanup is unchanged and remains the normal release path; the cap is only a backstop for sessions that never get one.

Evicting a checkpoint state is safe: the dedup-critical fields (`capturedFillBands`, `capturedQualityThresholds`, `capturedMilestones`, `lastCheckpointAt`) are written through to disk on every `recordCheckpointDecision()`, and `getCheckpointState()` re-hydrates an evicted entry from disk — so a band/threshold/milestone is never re-captured. The transient counters (`editWriteCount`, `editedFiles`, `lastEvaluatedAt`) are not persisted; evicting an uncheckpointed session resets them, which only *delays* a milestone-edit-batch checkpoint (conservative, never spurious). These trade-offs are documented in the code.

## How (verified)

Verified the eviction algorithm: inserting 2500 sessions caps size at 2000, evicts the oldest, retains the newest, and a refreshed key survives a later eviction wave. `tsc` build is green; committed `dist/` rebuilt.

## Deliberately not addressed

The cap is hardcoded (not a new `TOKEN_OPTIMIZER_*` env knob) and the eviction helper is not extracted to a shared module — both kept out to keep this a minimal, focused backstop rather than an API/structure change.


alexgreensh:
Merged in v5.11.13 — good catch on the map growth when sessions end uncleanly. Thanks, Dani.

PR #65, opened 2026-06-15
harden(checkpoints): refuse to follow symlinks on checkpoint writes (O_NOFOLLOW)

danikdanik:
## Why

The checkpoint write paths (`smart-compact.ts`) and the pending-hint write (`continuity.ts`) `lstat` their target before writing, but a TOCTOU window remains between that check and the write — another process could swap the target for a symlink in between. The existing `lstat`/`realpath` guards are therefore defense-in-depth, not provably safe, if the checkpoint root's parent is shared/world-writable.

## What

Add a `writeFileNoFollow()` helper that opens the target with `O_NOFOLLOW`, so the kernel rejects the open (`ELOOP`) when the final path component is a symlink at open time, then writes through the fd. Applied to both checkpoint body writes, the manifest append (`O_APPEND | O_NOFOLLOW`), and the continuity pending-hint write. Existing regular files are still overwritten in place (`O_TRUNC`).

## Scope / limitations (called out in the code — this narrows the race, it does not close it)

- `O_NOFOLLOW` protects only the **final** path component, not parent directories (those remain covered by the existing `lstat`/`realpath` checks).
- It rejects **symlinks** but **not hardlinks**; a pre-planted fixed-name hardlink would need a temp-file-then-rename pattern, out of scope for this symlink fix.
- **POSIX-only:** on Windows `fs.constants.O_NOFOLLOW` is `undefined` and ORs in as `0`, so the write still succeeds but retains the pre-existing `lstat`-only protection (no regression).

The helper is duplicated in both files deliberately, matching the existing convention in `continuity.ts` (which already re-implements the path-safety helpers locally rather than coupling to `checkpoint-policy.ts`).

## How (verified)

Writing through a symlink now throws `ELOOP` and leaves the link target untouched, while normal new-file writes and in-place overwrites still succeed. `tsc` build is green; committed `dist/` rebuilt.


alexgreensh:
Merged in v5.11.13. Solid hardening. As follow-up I extended O_NOFOLLOW to two more write paths in checkpoint-policy.ts (persistState, appendCheckpointEvent) and pulled the helper into a shared fs-utils.ts. Thanks!

PR #66, opened 2026-06-15
perf(session-parser): read each session file once in extractCostlyPrompts

danikdanik:
## Why

`extractCostlyPrompts()` read the same session `.jsonl` file from disk **twice** on every call: once inside `parseSessionTurns()` (which it calls for per-turn cost data) and once more directly, to walk the file for user-message text. Session transcripts can be multiple megabytes, and the audit / dashboard path calls `extractCostlyPrompts()` once per costly session, so each call paid for two full `readFileSync` + two `split("\n")` passes over the same bytes.

## What

`extractCostlyPrompts()` now reads and splits the file **once** and reuses the resulting lines for both the per-turn parse and the user-text walk. Net effect per call: one `statSync` + one `readFileSync` + one `split` instead of two of each. As a side benefit, pairing user text to turns is now immune to the file changing between the two former reads (a small time-of-check window is gone).

## How

- Extracted a private `parseSessionTurnsFromLines(lines, openclawDir)` core holding the existing turn-parsing logic verbatim.
- `parseSessionTurns(filePath, ...)` now reads + splits, then delegates to that core. Its public signature and behavior are unchanged.
- `extractCostlyPrompts()` reads + splits once, calls the core for turns, then walks the same `lines` array for user text. Because both walks operate on the identical `lines`, the turn-index pairing stays exactly aligned (the constraint already flagged by the in-code comment).

## Validation

- `npm run build` (tsc) passes; compiled `dist/` regenerated and committed in lockstep with `src` (repo convention).
- Behavior verified **byte-identical** before/after on a fixture exercising normal turns, a sidechain message, a tool_result-only user message, and a trailing user message (the turn-index-alignment-sensitive cases).
- Public API surface unchanged: no `.d.ts` signature change; the new helper is private.

## Scope

Deliberately narrow. Left for a possible follow-up: the broader audit pipeline still reads each file independently via `parseSession`, `extractCostlyPrompts`, and `loadMessagesFromSessionFile`; and the `statSync` / read / 50MB-guard block is duplicated across the two functions.


alexgreensh:
Merged in v5.11.13 — single read plus the turn-pairing-stability bonus. Thanks, Dani.

PR #67, opened 2026-06-15
perf(pricing): memoize normalizeModelName

danikdanik:
## Why

`normalizeModelName()` maps a model id (e.g. `anthropic/claude-sonnet-4-...`) to a pricing key by running up to ~60 sequential `String.includes()` checks (Anthropic -> GPT-5 family -> GPT-4 -> reasoning -> Gemini -> DeepSeek -> Qwen -> ...). It sits on hot paths and is called many thousands of times per audit:

- once per usage bucket in `parseSession`, and again inside `calculateCost`;
- once per turn in `parseSessionTurns`;
- and across the savings / quality / dashboard accumulation loops.

Almost every call is on a small, repeating set of model-id strings, so the same ~60-branch scan is redone over and over.

## What

Memoize the result. `normalizeModelName()` now consults a module-level `Map` cache keyed on the raw model id; repeat lookups become O(1) instead of a linear scan. The matching logic is unchanged.

## How

- Extracted the existing matching logic **verbatim** into a private `computeNormalizedModelName(modelId)`.
- `normalizeModelName()` checks the cache with `Map.has()` (not a falsy check, so a cached `null` — for empty or `<...>` ids — is distinguished from a miss), computes on a miss, and stores the result.
- The cache is **bounded** (cap 4096; once full, new ids are still computed correctly, just not cached) so a pathological stream of unique ids can't grow it unbounded. The realistic key space is a few dozen ids, so the cap is never approached.
- Wired `_normalizeCache.clear()` into the existing `resetPricingCache()`, mirroring the file's existing `_mergedPricing` cache idiom.

## Validation

- `npm run build` (tsc) passes; `dist/` regenerated and committed in lockstep with `src`.
- Output verified **byte-identical** before/after across ~70 model strings: provider prefixes (`anthropic/`, `openrouter/openai/`, `anthropic:`), mixed case, every GPT-5 / GPT-4 / Gemini / DeepSeek / Qwen / Mistral / Grok / local variant, and null-producing inputs (`""`, `"<...>"`).
- Idempotency confirmed (repeat call == first call) and `resetPricingCache()` confirmed to clear the cache without changing results.
- `normalizeModelName` is a pure function of its argument (no dependency on pricing config or env), so caching is safe regardless of config reload.
- Public API surface unchanged: no `.d.ts` signature change; `computeNormalizedModelName` is private.

## Scope

Narrow by design. A parallel `normalizeModelName` exists in `opencode/src/pricing.ts` without this cache; mirroring it there is a possible follow-up (lower urgency — OpenCode reprices per session, not per turn).


alexgreensh:
Merged in v5.11.13 — clean null-aware memoization (nice use of .has() over a falsy check). Thanks!

PR #68, opened 2026-06-16
feat(archive): full-file path in pointer + env-configurable knock-in/knock-out

niemst:
## What
- **Archived-result message now includes the on-disk file path** (plus the `expand` id), so a full result is directly retrievable — not only via `expand <key>`.
- **Knock-in / knock-out thresholds are now env-configurable:**
  - `TOKEN_OPTIMIZER_ARCHIVE_THRESHOLD` — knock-in **X** (min size to archive)
  - `TOKEN_OPTIMIZER_ARCHIVE_PREVIEW_SIZE` — knock-out **Y** (inline preview cap)
- **Invariant X > Y enforced**: if misconfigured (Y ≥ X), Y is clamped just below X rather than crashing the hot-path hook.
- Loaded the literal `1500` preview size into the named constant; `build_archive_pointer` gains an optional, backward-compatible `archive_path` arg.

## Notes
- **Defaults unchanged (4096 / 1500)** — behavior is identical unless the env vars are set.
- Both mirrored copies (`skills/…` and `plugins/token-optimizer/skills/…`) updated identically.
- `py_compile` clean; verified default / override / clamp / invalid-fallback at import time.

### DIFF
--- a/plugins/token-optimizer/skills/token-optimizer/scripts/archive_result.py
+++ b/plugins/token-optimizer/skills/token-optimizer/scripts/archive_result.py
@@ -50,8 +50,31 @@ try:
     from token_estimate import CODE_CHARS_PER_TOKEN
 except Exception:  # pragma: no cover - fail-open if shared estimator missing
     CODE_CHARS_PER_TOKEN = 3.3
-_ARCHIVE_THRESHOLD = 4096       # chars: only archive results >= this size
-_ARCHIVE_PREVIEW_SIZE = 1500    # chars: preview included in replacement output
+def _positive_int_env(name: str, default: int) -> int:
+    """Read a positive-int override from the environment; fall back to ``default``.
+
+    Never raises — invalid or non-positive values are ignored so a misconfigured
+    env var can never crash this hot-path hook.
+    """
+    raw = os.environ.get(name, "").strip()
+    if raw:
+        try:
+            value = int(raw)
+            if value > 0:
+                return value
+        except ValueError:
+            pass
+    return default
+
+
+# Compression hysteresis (knock-in / knock-out). A result is only archived once it
+# reaches the knock-in threshold X; when archived, the inline preview is capped at
+# the knock-out size Y. Both are env-tunable. Invariant X > Y guarantees compressing
+# always yields a real saving; if misconfigured, Y is clamped below X (never crashes).
+_ARCHIVE_THRESHOLD = _positive_int_env("TOKEN_OPTIMIZER_ARCHIVE_THRESHOLD", 4096)        # knock-in (X)
+_ARCHIVE_PREVIEW_SIZE = _positive_int_env("TOKEN_OPTIMIZER_ARCHIVE_PREVIEW_SIZE", 1500)  # knock-out (Y)
+if _ARCHIVE_PREVIEW_SIZE >= _ARCHIVE_THRESHOLD:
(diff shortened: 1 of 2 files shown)


alexgreensh:
Thanks for digging into the archive path, @niemst — appreciate the careful defensive coding (the `_positive_int_env` clamp is genuinely tidy). After review I'm going to pass on this one as-is, and I want to be transparent about why since the two halves landed together:

**Env-configurable thresholds — declining.** The `4096`/`1500` constants aren't general config; they're the compression heuristic's core dial, validated against our harm/savings gate. Exposing them as knobs lets a setup silently drift net-negative with no gate re-validation, which cuts against Token Optimizer's "sensible defaults, no knob sprawl" design. There's also a correctness snag: `TOKEN_OPTIMIZER_ARCHIVE_PREVIEW_SIZE` only resizes the internal telemetry preview, not the model-visible replacement preview, so the knob doesn't do what its name promises.

**File path in the archived message — good idea, but I'd land it differently.** Surfacing a raw path as a fallback for harnesses where `expand` isn't wired is worth having. But as written it only updates the inline MCP branch — `build_archive_pointer` gained the `archive_path` param while its actual caller (`read_cache.py:883`) was left on the 3-arg form, so the most common archive path (Read skeletons) gets nothing. Rather than have you chase the wiring across both mirrored trees, we'll fold this into our own pass and credit the idea.

Heads-up separately: the CLA check is still red because the sign-off comment hasn't been posted — unrelated to the review above, just flagging so it's not a mystery.

Thanks again for the contribution and the clear write-up.

alexgreensh:
Thanks @niemst, this is a nice quality-of-life change. Surfacing the on-disk path in the archive pointer (not just the `expand` id) and making the knock-in/knock-out thresholds env-configurable both make sense, and I like that the defaults are unchanged and the X > Y invariant is clamped rather than crashing the hot path.

One thing before it can merge: the CLA still needs signing. Good news is the CLA bot itself was just fixed (it wasn't able to clear its own check after signing, which is probably why earlier signers saw it stay red). So signing now should actually turn the check green. To sign, post a comment with exactly:

Once that's in, I'll give it a proper review pass and get it merged. Appreciate the contribution.

niemst:

niemst:
Apologies for the delay; I've been swamped with work

alexgreensh:
Thank you, @niemst 🙏 Going to close this one, and I want to be straight about why.

The **path-in-pointer** half is a good idea and we are keeping it. Surfacing the on-disk path as a fallback for harnesses where `expand` is not wired is genuinely worth having, and I am folding it into our own pass so it goes through the actual caller (`read_cache.py` was still on the 3-arg `build_archive_pointer`, so the most common Read-skeleton path got nothing). Full credit to you for the idea.

The **env-configurable thresholds** I am passing on. The `4096` / `1500` constants are the compression heuristic's core dial, validated against our harm/savings gate; exposing them as knobs lets a setup drift net-negative with no re-validation, which cuts against the project's "sensible defaults, no knob sprawl" line. There is also a naming snag: `TOKEN_OPTIMIZER_ARCHIVE_PREVIEW_SIZE` only resizes the internal telemetry preview, not the model-visible replacement preview.

Really appreciate the careful defensive coding (the `_positive_int_env` clamp is tidy) and the clear write-up. Thank you for contributing, and please do send more. 🙌

PR #69, opened 2026-06-16
fix(runtime): honor CLAUDE_CONFIG_DIR instead of hardcoding ~/.claude

neilinger:
`claude_home()` and a few scripts/hooks hardcoded `~/.claude` and ignored `CLAUDE_CONFIG_DIR` — Claude Code's official config-dir override. Every other runtime already honors its home env var via `_safe_home_from_env`; `claude_home()` was the lone exception. With a relocated config dir, this pins session scans, the trends DB, the dashboard, and config/hook paths to a stale `~/.claude`.

Routes `claude_home()` through the existing `_safe_home_from_env` helper (same safe-home confinement + symlink rejection as the other runtimes), and the direct callers through `claude_home()` (`run.py` honors it inline). Falls back to `~/.claude` when unset. Applied to both mirrored source trees (`skills/` and `plugins/`).

Files: `runtime_env.py` (keystone), `hooks/run.py`, `compression_backfill.py`, `archive_result.py`, `structure_replay.py`, `hermes_hook_bridge.py`.

Note: `_safe_home_from_env` confines the override to a non-symlink dir under `$HOME`; a `CLAUDE_CONFIG_DIR` outside `$HOME` is rejected with a warning and falls back. Kept for consistency with the other runtimes — happy to relax if you'd prefer full Claude-Code semantics.

### DIFF
--- a/hooks/run.py
+++ b/hooks/run.py
@@ -52,7 +52,16 @@ def _check_consent() -> bool:
                     return True
                 config_path = ch / "token-optimizer" / "config.json"
             else:
-                config_path = home / ".claude" / "token-optimizer" / "config.json"
+                # Honor CLAUDE_CONFIG_DIR (Claude Code's official config-dir
+                # override) before falling back to ~/.claude.
+                claude_config = os.environ.get("CLAUDE_CONFIG_DIR", "")
+                if claude_config:
+                    cc = Path(claude_config).resolve()
+                    if not str(cc).startswith(str(home)):
+                        return True  # Path outside home = skip (fail-open)
+                    config_path = cc / "token-optimizer" / "config.json"
+                else:
+                    config_path = home / ".claude" / "token-optimizer" / "config.json"
 
         if not config_path.exists() or config_path.is_symlink():
             return True  # No config or symlink = fail-open
--- a/plugins/token-optimizer/hooks/run.py
+++ b/plugins/token-optimizer/hooks/run.py
@@ -52,7 +52,16 @@ def _check_consent() -> bool:
                     return True
                 config_path = ch / "token-optimizer" / "config.json"
             else:
-                config_path = home / ".claude" / "token-optimizer" / "config.json"
+                # Honor CLAUDE_CONFIG_DIR (Claude Code's official config-dir
+                # override) before falling back to ~/.claude.
+                claude_config = os.environ.get("CLAUDE_CONFIG_DIR", "")
+                if claude_config:
+                    cc = Path(claude_config).resolve()
+                    if not str(cc).startswith(str(home)):
+                        return True  # Path outside home = skip (fail-open)
+                    config_path = cc / "token-optimizer" / "config.json"
+                else:
+                    config_path = home / ".claude" / "token-optimizer" / "config.json"
 
         if not config_path.exists() or config_path.is_symlink():
             return True  # No config or symlink = fail-open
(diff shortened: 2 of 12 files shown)


alexgreensh (changes requested):
Thanks for this, @neilinger — you caught a real one. `claude_home()` was genuinely the lone runtime hardcoding `~/.claude` while every sibling honors its home env, and the stale-dir failure mode is exactly as you describe. I want to merge it. Two changes first:

**1. Honor `CLAUDE_CONFIG_DIR` outside `$HOME`.** Claude Code permits `CLAUDE_CONFIG_DIR` to point anywhere (containers, CI runners, non-home `HOME`, relocated config volumes). Routing through `_safe_home_from_env` rejects those and silently falls back to `~/.claude` — which reintroduces the exact stale-dir bug this PR fixes, just for a different slice of users, with only a swallowed stderr warning. You flagged this trade-off yourself in the description, and I'd like to take you up on relaxing it: a resolver that accepts any absolute, existing, non-symlink directory (keep the symlink + relative-path rejection for traversal safety, drop the under-`$HOME` confinement). `CLAUDE_CONFIG_DIR` has broader semantics than `CODEX_HOME`/`HERMES_HOME`, so it warrants its own resolver rather than the shared one.

**2. Align the `hooks/run.py` inline check.** The consent gate hand-rolls its own `CLAUDE_CONFIG_DIR` resolution that (a) diverges from `claude_home()` for out-of-home and not-yet-created dirs, and (b) uses `str(cc).startswith(str(home))`, which sibling-prefix-matches (`/Users/alex-evil` passes for `/Users/alex`). Please switch it to `is_relative_to(home)` at minimum, ideally to the same resolver, so consent and data resolution agree on one directory.

A small regression test asserting `claude_home()` honors a valid `CLAUDE_CONFIG_DIR` (in- and out-of-home) and rejects symlinks would seal it. Appreciate the clean dual-tree mirroring and the keystone routing — this is 90% there.

alexgreensh:
Hi @neilinger, still keen on this one. You caught a real bug (`claude_home()` was the lone runtime hardcoding `~/.claude`) and the fix is most of the way there.

Two quick heads-ups so it can land:

1. It'll need a small rebase onto `main`. I just landed some adjacent UTF-8/locale hardening that also touches `hooks/run.py` and `runtime_env.py`, so there may be light conflicts near those.
2. The two review notes from before still apply: let `CLAUDE_CONFIG_DIR` point outside `$HOME` (a dedicated resolver that keeps the symlink + relative-path rejection but drops the under-$HOME confinement, since Claude Code allows it anywhere), and align the inline `run.py` consent check to use `is_relative_to(home)` rather than `startswith`.

No rush at all, but if you're heads-down on other things and would rather I carry it the last 10% (rebase + those two tweaks + a small regression test), just say the word and I'm happy to. Thanks again for the contribution.

alexgreensh:
Landed in **v5.11.17**. Thank you, @neilinger 🙏

Your commit is in `main` (cherry-picked, your authorship preserved), and I folded the two review asks in on top:

1. `claude_home()` now accepts `CLAUDE_CONFIG_DIR` **outside `$HOME`** (containers, CI runners, relocated config volumes), while keeping the symlink and relative-path rejection for traversal safety. It gets its own resolver rather than the `$HOME`-confined `_safe_home_from_env` (which stays correct for `CODEX_HOME` / `HERMES_HOME` / `COPILOT_HOME`).
2. The `hooks/run.py` consent check dropped the hand-rolled `str(cc).startswith($HOME)` match (it sibling-prefix-matched, e.g. `/Users/alex-evil` passing for `/Users/alex`) in favor of the same acceptance logic.

Plus `tests/test_claude_config_dir.py` covering unset, in-home, out-of-home, symlink, relative, nonexistent, and empty.

You caught a genuine one: the lone runtime that hardcoded `~/.claude` while every sibling honored its home env. Closing this PR since it has landed; the credit is in the git history and the release notes. Thanks for the clean dual-tree mirroring and the clear write-up. 🙌

PR #72, opened 2026-06-22
audit: fix stale hardcoded date in backfill TTL test

glupta:
## Summary

- `test_backfill_cache_ttl_mix_updates_existing_rows` hardcoded `date = "2026-04-04"`, which is now > 30 days in the past
- `_backfill_session_metrics` filters `WHERE date >= cutoff` (today − 30 days ≈ 2026-05-23), so the row was silently skipped and `updated == 0 != 1`
- Fix: replace the literal with `datetime.now().strftime("%Y-%m-%d")` so the test row is always within the backfill window
- Also adds `sys.path` preamble so `measure.py`'s local imports (`hook_io`, `plugin_env`, etc.) resolve without an editable install
- Removes `tests/` from `.gitignore` so the test suite can be tracked

## Test plan

- [x] `python3 -m pytest tests/test_cache_ttl_mix.py -v` — all 7 tests pass
- [x] Confirmed `test_backfill_cache_ttl_mix_updates_existing_rows` was previously `FAILED: AssertionError: 0 != 1` and is now `PASSED`

STEWARD-RISK: low
STEWARD-CLASS: test

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,7 +19,6 @@ node_modules/
 venv/
 snapshots/
 .pytest_cache/
-tests/
 FIELD_NOTES.md
 compound-engineering.local.md
 private-review/
--- a/tests/test_cache_ttl_mix.py
+++ b/tests/test_cache_ttl_mix.py
@@ -1,0 +1,285 @@
+import importlib.util
+import json
+import sqlite3
+import tempfile
+import unittest
+from datetime import datetime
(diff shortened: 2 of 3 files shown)


glupta:
Re-targeting to my fork glupta/token-optimizer — these audit tests belong on the fork, not as an unsolicited upstream PR.

PR #74, opened 2026-06-24
feat(detector): respondToBashCommands settings check

danikdanik:
## Why

[Claude Code v2.1.186](https://github.com/anthropics/claude-code/releases/tag/v2.1.186) introduced a new default: every `! bash` command output now triggers an automatic model reply (`respondToBashCommands: true`). From the release notes:

> *"! bash commands now trigger Claude to respond to the output automatically; set `respondToBashCommands: false` in settings.json to keep the previous context-only behavior."*

This behavior is invisible - there's no warning, and the token cost rolls into session totals with no attribution. Users who haven't explicitly disabled it are silently spending output tokens on replies they didn't ask for. The fix is a one-liner in `settings.json`, but only if you know the setting exists.

## What

New `respond_to_bash_commands` detector that checks `~/.claude/settings.json` directly. It fires whenever `respondToBashCommands` is not explicitly `false`, with a suggestion pointing to the exact file and key to change.

Fires in these cases:
- Key is absent (default state - most users upgrading past v2.1.186)
- Key is `true` or any truthy value
- `settings.json` doesn't exist yet (same as absent)
- File contains invalid JSON (treated as empty settings)

Does not fire when `respondToBashCommands: false` is explicitly set.

## How

**Settings-check pattern** (not session JSONL parsing). The setting is a one-time config fact, not a per-session signal, so parsing turns would be the wrong layer.

Key implementation decisions:

- **Missing file fires the finding** - a user with no `settings.json` has definitely never set the key; suppressing the finding here would silently miss the most common case
- **`always_show: true`** in the finding dict - bypasses the registry's `confidence > 0.3` gate, which is designed for session-derived signals, not static config checks
- **Name-based dedup in `run_all_detectors`** - prevents the same config finding from appearing N times in a batch run over multiple session files
- **`runtime_env.claude_home()`** honored at module load (ImportError-guarded, falls back to `Path.home() / ".claude"`) - `CLAUDE_CONFIG_DIR` users get the right path in both the check and the suggestion text
- **PermissionError** logs to stderr and returns `[]` (can't confirm the setting; don't false-positive)

## Testing

13 tests covering all branches: file absent, key missing, key true, key false, PermissionError, invalid JSON, JSON null, `always_show` bypass, registry dedup, sentinel thresholds, `runtime_env` fallback via `sys.modules`.

### DIFF
--- a/plugins/token-optimizer/skills/token-optimizer/scripts/detectors/registry.py
+++ b/plugins/token-optimizer/skills/token-optimizer/scripts/detectors/registry.py
@@ -1,4 +1,4 @@
-"""Detector registry: session-level waste detectors with run helper."""
+"""Detector registry: session and config detectors with run helper."""
 
 from detectors.retry_churn import detect_retry_churn
 from detectors.tool_cascade import detect_tool_cascade
@@ -28,7 +28,7 @@ _TRIAGE_MIN_TOKENS = 5000
 
 
 def run_all_detectors(session_data):
-    """Run all session-level detectors. Returns sorted, deduplicated findings list."""
+    """Run all session and config detectors. Returns sorted findings list."""
     findings = []
     for d in ALL_DETECTORS:
         try:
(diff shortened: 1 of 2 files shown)


alexgreensh:
Thank you so much for this, @danikdanik! 🙏 This is a genuinely useful detector and the care you put in really shows — 13 test cases, proper `runtime_env` import guard with a fallback, `PermissionError` handled as "unknown" so it never false-fires, and a clean `always_show` plumbing through `triage()` and `run_all_detectors()`. Lovely work.

I've **shipped it in v5.11.20** with full credit to you. 🎉 Two small notes on what changed during integration, just so you can see the reasoning:

1. **Moved to the canonical tree.** Your edits landed in `plugins/token-optimizer/`, which is auto-generated from `skills/` by `scripts/sync-codex-marketplace-plugin.sh` (and rebuilt on every release), so a direct merge would've been overwritten on the next sync. Totally non-obvious from the outside — the repo layout doesn't advertise it. I ported `respond_to_bash.py` + the registry change into `skills/token-optimizer/scripts/detectors/` so it sticks.

2. **Made the `always_show` finding actually reach users.** `triage()` and `run_all_detectors()` honored `always_show` perfectly, but `measure.py`'s display had two independent gates the flag didn't know about: an unguarded `f['savings_tokens']` access (KeyError for a config finding) and a 5%-message-frequency noise gate that suppressed it for any session over ~20 messages. Added `savings_tokens: 0` to the finding and taught both gates to respect `always_show`, so it now renders as a clean "Config fix".

Closing as merged-in-spirit and shipped in [v5.11.20](https://github.com/alexgreensh/token-optimizer/releases/tag/v5.11.20). Really appreciate the contribution — please send more our way! 🚀

alexgreensh:
Shipped in v5.11.20 with credit — thank you again, @danikdanik! 🙏

PR #84, opened 2026-07-07
Redact credentials passed as URL query parameters

nointerview1548:
## Problem

The credential redactor in `credential_patterns.py` ships 22 patterns (Bearer tokens, key-shaped strings, DB URIs, HTTP basic-auth URLs, etc.), but none of them match a credential passed as a **URL query parameter**. Tool output frequently contains such URLs, and `archive_result.py` archives it verbatim after redaction, so a query-param secret lands unredacted in both the tool-archive JSON and the SQLite preview.

### Repro (before this change)

```python
from credential_patterns import redact_credentials
redact_credentials("GET https://example.com/alerts?token=SECRETVALUE HTTP/1.1")
# -> 'GET https://example.com/alerts?token=SECRETVALUE HTTP/1.1'   (unchanged!)
```

The same holds for `?api_key=`, `?access_token=`, `?auth=`, etc. Because `archive_result._redact_credentials` delegates to this function, the value is written to the archive as-is.

## Fix

Add a 23rd pattern for the URL auth-param class, matching the common credential-bearing parameter names case-insensitively:

```
?token= | ?key= | ?api_key= | ?api-key= | ?access_token= | ?auth=
```

A named `keep` group captures the `?name=` prefix so redaction **preserves the parameter name and blanks only the value**, and the value match stops at the next `&`, whitespace, or quote so any following parameters survive:

```python
redact_credentials("https://example.com/a?access_token=SECRET&page=2")
# -> 'https://example.com/a?access_token=[CREDENTIAL REDACTED: URL auth param]&page=2'
```

`redact_credentials` now honors a `keep` group generically; **the existing 22 patterns are unchanged** (they have no `keep` group and continue to redact the whole match). This complements the existing redactor rather than altering it, and stays within `SECURITY.md`'s redact-before-write model (the value is blanked before it is ever persisted).

## Tests

Adds `tests/test_credential_patterns.py` (force-added past the `tests/` gitignore rule, matching the existing test files) covering:

- value redaction across `token` / `api_key` / `api-key` / `access_token` / `auth` / `key`
- parameter-name preservation (only the value is blanked)
- following params surviving (`&page=2&sort=asc` is not swallowed)
- **negative cases** that must NOT be redacted: `?page=2`, `?monkey=1` (contains `key` only as a substring), `?tokenizer=fast` (`token` only as a name prefix), and query values that merely contain the words `token`/`auth`
- a regression check that representative existing patterns still fully redact

All fixtures are fake (e.g. `FAKE_URL_TOKEN_123`); no real secret appears anywhere.

```
$ python3 -m pytest tests/ -v
53 passed
```

(39 pre-existing tests + 14 new; the full suite is green.)

## Notes

- Single file of production change (`credential_patterns.py`), 16 insertions / 2 changed lines; no change to the public `CREDENTIAL_PATTERNS` tuple shape, so `measure.py`, `bash_compress.py`, and `read_cache.py` consumers are unaffected (the security-report credential count simply goes 22 -> 23).
- I did not touch `CHECKSUMS.sha256`, since that is regenerated at release time.

### DIFF
--- a/skills/token-optimizer/scripts/credential_patterns.py
+++ b/skills/token-optimizer/scripts/credential_patterns.py
@@ -34,6 +34,11 @@ CREDENTIAL_PATTERNS: List[Tuple[str, "re.Pattern[str]"]] = [
     ("PEM private key",         re.compile(r"-----BEGIN [A-Z ]*PRIVATE KEY-----")),
     ("Database URI",            re.compile(r"(?:postgres|postgresql|mysql|mongodb|mongodb\+srv|redis)://[^:\s/]+:[^@\s]+@", re.I)),
     ("HTTP basic auth URL",     re.compile(r"https?://[^:\s/@]+:[^@\s]+@", re.I)),
+    # Credentials passed as URL query parameters (e.g. ?token=..., ?api_key=...).
+    # The named `keep` group captures the "?name=" prefix so redaction preserves the
+    # parameter name and blanks only the value (see redact_credentials). The value
+    # match stops at the next `&`, whitespace, or quote so following params survive.
+    ("URL auth param",          re.compile(r"(?P<keep>[?&](?:token|key|api[_-]?key|access[_-]?token|auth)=)[^&\s\"']+", re.I)),
 ]
 
 # Bare compiled patterns list for backward compat with bash_compress.py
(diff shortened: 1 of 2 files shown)


nointerview1548:

PR #86, opened 2026-07-07
perf: mark 11 stdout-free hooks async, split archive_result.py matcher

danikdanik:
## Problem

`hooks.json` has 24 hook command entries. Only 2 are marked `"async": true` (`ensure-health`, `session-end-flush`). Every other hook blocks its event even when nothing it does is ever read back — most of them just write a cache/checkpoint file to disk.

## Why this isn't a blanket flip

Per Claude Code's hook docs, **async hooks are strictly fire-and-forget**: stdout and JSON output are completely discarded, and an async hook cannot block a tool call, set `permissionDecision`, or inject `additionalContext`. So a hook is only safe to mark async if its entire job is a side effect nobody consumes — flipping the wrong one silently breaks whatever it was supposed to inject or block.

I checked each candidate's actual output contract in source rather than going by name.

**Marked async** (confirmed silent under `--quiet`, or the only output path is gated on a flag this specific call never passes):

| hook | event(s) | why safe |
|---|---|---|
| `checkpoint-trigger` | PreToolUse `Agent\|Task` | its only `print` is gated `if not quiet` |
| `compact-capture --trigger auto` | PreCompact only | confirmed silent under `--quiet`; the Stop/StopFailure instances of this same subcommand are **not** touched, see below |
| `read_cache.py --clear` | PreCompact, CwdChanged | stderr-only |
| `read_cache.py --invalidate` | PostToolUse `Edit\|Write\|MultiEdit\|NotebookEdit` | stderr-only |
| `quality-cache --force` | SessionStart, PostCompact | no `--warn` passed → the warn-emission branch is unreachable |
| `quality-cache --quiet --throttle-only` | PostToolUse catch-all (broadest matcher: `Bash\|Read\|Glob\|Grep\|Agent\|Edit\|Write\|MultiEdit\|NotebookEdit\|mcp__.*`) | same reasoning as above — no `--warn` flag on this call, so the warn branch is dead code here specifically |
| `keepwarm-arm` | Stop | docstring says "additive", explicitly fail-open by design |
| `context_intel.py` | PostToolUse | zero stdout prints; the file's own docstring says "no additionalContext, zero cache impact" |
| `archive_result.py` | PostToolUse, **only** for matcher `Bash\|Read\|Glob\|Grep\|Agent` | see split below |

**Left sync** (must not change):

- `read_cache.py` on PreToolUse/Read, `bash_hook.py` on PreToolUse/Bash — both emit `permissionDecision` / cache-hit `additionalContext`. This is the actual read-caching mechanism.
- `dynamic-compact-instructions`, both `compact-restore` variants — the compaction-survival mechanism itself.
- `compact-capture` on **Stop and StopFailure specifically** — its stdout is silent, but the risk here isn't output, it's completion. The hook docs don't guarantee an async child process survives the parent exiting right after `Stop`, and killing this mid-write would corrupt the checkpoint the plugin's "survives compaction" promise depends on. Kept conservative.
- `UserPromptSubmit`'s three hooks: `quality-cache --warn` (confirmed — this is exactly the warning-injection path the throttle-only variant above lacks), `prompt-continuity`, `verbosity-steer` (didn't fully trace these two, treated conservatively as sync rather than guess).
- `archive_result.py`'s `mcp__.*` case — this is the one that had to be split out, see below.

## Measured impact

Async doesn't make the underlying work faster (that's PR #85) — it removes it from the *blocking* path entirely: Claude Code doesn't wait for an async hook before continuing, and doesn't consume its output. So the number that matters here is how much blocking time these hooks were adding per event, measured with the full `python-launcher.sh` → `run.py` → target chain, on `main` (i.e. before PR #85's dispatch fix too, so this is the current, un-optimized cost being removed):

| chain | hooks involved | blocking time removed |
|---|---|---|
| PostToolUse, per matching tool call (typical response, under the archive threshold) | `archive_result.py` + `context_intel.py` + `quality-cache --throttle-only`, run sequentially | 0.56-0.66s → 0s |
| PreCompact | `compact-capture --trigger auto` + `read_cache.py --clear` (the 3rd PreCompact hook, `dynamic-compact-instructions`, stays sync) | 0.82s → 0s |

The PostToolUse one is the one that matters most: that matcher fires on nearly every tool call in a session, so today every matching tool call pays that ~0.6s in the critical path before Claude Code can proceed. After this PR it pays 0s — the archiving/cache-refresh work still happens, just off to the side.

## The archive_result.py split

The existing single entry had matcher `Bash|Read|Glob|Grep|Agent|mcp__.*` covering both cases with one command. But `archive_result.py` branches internally: for MCP tools (`tool_name` containing `__`) it emits `updatedMCPToolOutput` — the actual token-compression replacement, i.e. the plugin's own value proposition for large MCP responses. Marking that async would silently let the full uncompressed MCP output through instead, since the replacement JSON would just get dropped. For every other tool type the same function only writes durability metadata to disk (its own docstring says as much), which is safe to make async.

So this PR splits the one entry into two matcher-scoped entries: `mcp__.*` stays sync (untouched), `Bash|Read|Glob|Grep|Agent` gets `"async": true`. Same command, same script — just no longer sharing one matcher across a sync-required case and an async-safe one.

## plugins/ mirror

`plugins/token-optimizer/` (the Codex marketplace mirror) regenerated via the repo's own `scripts/sync-codex-marketplace-plugin.sh`, which already strips `async` from the mirror (Codex doesn't support async hooks yet and would skip them entirely if left in — that's existing, documented behavior in the script, not something this PR changes). Confirmed the matcher split itself survives the strip correctly; only the `async` keys are removed there.

## Test plan

Added `tests/test_async_hook_wiring.py`, which pins the exact classification above: parses `hooks.json`, asserts every hook entry's `async` value matches the table (including the must-stay-sync ones), asserts the total async count, and asserts the mirror has every `async` key stripped but is otherwise structurally identical to root. This should make it hard for a future edit to silently flip an unsafe one.

Full test suite passes.

## References

- [Claude Code hooks reference](https://code.claude.com/docs/en/hooks) — async hook semantics: fire-and-forget, stdout/JSON discarded, cannot block or inject context.

### DIFF
--- a/hooks/hooks.json
+++ b/hooks/hooks.json
@@ -24,7 +24,8 @@
         "hooks": [
           {
             "type": "command",
-            "command": "for b in bash /bin/bash /usr/bin/bash /usr/local/bin/bash /opt/homebrew/bin/bash; do command -v \"$b\" >/dev/null 2>&1 && exec \"$b\" \"${CLAUDE_PLUGIN_ROOT}/hooks/python-launcher.sh\" \"${CLAUDE_PLUGIN_ROOT}/hooks/run.py\" skills/token-optimizer/scripts/measure.py checkpoint-trigger --milestone pre-fanout --quiet; done; exit 0"
+            "command": "for b in bash /bin/bash /usr/bin/bash /usr/local/bin/bash /opt/homebrew/bin/bash; do command -v \"$b\" >/dev/null 2>&1 && exec \"$b\" \"${CLAUDE_PLUGIN_ROOT}/hooks/python-launcher.sh\" \"${CLAUDE_PLUGIN_ROOT}/hooks/run.py\" skills/token-optimizer/scripts/measure.py checkpoint-trigger --milestone pre-fanout --quiet; done; exit 0",
+            "async": true
           }
         ]
       }
@@ -42,7 +43,8 @@
         "hooks": [
           {
             "type": "command",
-            "command": "for b in bash /bin/bash /usr/bin/bash /usr/local/bin/bash /opt/homebrew/bin/bash; do command -v \"$b\" >/dev/null 2>&1 && exec \"$b\" \"${CLAUDE_PLUGIN_ROOT}/hooks/python-launcher.sh\" \"${CLAUDE_PLUGIN_ROOT}/hooks/run.py\" skills/token-optimizer/scripts/measure.py compact-capture --trigger auto --quiet; done; exit 0"
+            "command": "for b in bash /bin/bash /usr/bin/bash /usr/local/bin/bash /opt/homebrew/bin/bash; do command -v \"$b\" >/dev/null 2>&1 && exec \"$b\" \"${CLAUDE_PLUGIN_ROOT}/hooks/python-launcher.sh\" \"${CLAUDE_PLUGIN_ROOT}/hooks/run.py\" skills/token-optimizer/scripts/measure.py compact-capture --trigger auto --quiet; done; exit 0",
+            "async": true
           }
         ]
       },
@@ -50,7 +52,8 @@
         "hooks": [
           {
             "type": "command",
-            "command": "for b in bash /bin/bash /usr/bin/bash /usr/local/bin/bash /opt/homebrew/bin/bash; do command -v \"$b\" >/dev/null 2>&1 && exec \"$b\" \"${CLAUDE_PLUGIN_ROOT}/hooks/python-launcher.sh\" \"${CLAUDE_PLUGIN_ROOT}/hooks/run.py\" skills/token-optimizer/scripts/read_cache.py --clear --quiet; done; exit 0"
+            "command": "for b in bash /bin/bash /usr/bin/bash /usr/local/bin/bash /opt/homebrew/bin/bash; do command -v \"$b\" >/dev/null 2>&1 && exec \"$b\" \"${CLAUDE_PLUGIN_ROOT}/hooks/python-launcher.sh\" \"${CLAUDE_PLUGIN_ROOT}/hooks/run.py\" skills/token-optimizer/scripts/read_cache.py --clear --quiet; done; exit 0",
+            "async": true
           }
         ]
       }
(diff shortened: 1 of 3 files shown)


alexgreensh:
Thank you for this, @danikdanik — genuinely careful work: a correctly diagnosed fire-and-forget contract, per-hook source verification instead of going by name, and a regression test pinning the exact wiring. Ten of the eleven flips plus the `archive_result.py` matcher split were verified correct on independent review and adversarial QA, and they've now landed on `main` (v5.11.38).

I applied a **reduced, verified-safe subset** rather than the full eleven. Async is now on: `checkpoint-trigger`, `compact-capture --trigger auto` (PreCompact), `archive_result.py` (the non-mcp side of your split — `mcp__.*` stays sync as you designed), and `context_intel.py`. That's the PostToolUse hot path, which is where the blocking cost actually lived.

Seven flips were reverted to sync after review + torture found real hazards:

- **`quality-cache --force` / `--throttle-only`** (SessionStart, PostCompact, PostToolUse): `quality_cache()` has a *second*, unconditional `systemMessage` stdout path (measure.py ~27982) that isn't gated by `--quiet`/`--warn` — the `--warn`-gated print you correctly identified as dead code is not the only emitter. Worse, it persists one-shot "already warned" dedup flags in the same call, so an async invocation that computes a warning marks it fired and Claude Code discards it — and then the still-sync `UserPromptSubmit --warn` fallback sees the flag set and never re-emits. The warning (including the once-per-session fresh-session nudge) is permanently lost, not delayed.
- **`read_cache.py --invalidate` / `--clear`**: same-session read-after-write race against the sync `PreToolUse/Read` cache reader on the same on-disk store.
- **`keepwarm-arm`** (Stop): kept sync for the same process-exit corruption-safety reason its Stop siblings are sync.

Your `tests/test_async_hook_wiring.py` was adapted to pin this classification (total async = 6, mirror-stripped parity preserved), with credit to you in the file header. Since the work is incorporated on `main`, I'm closing this PR — but the design and the test scaffold are yours, and they made the safe subset easy to land. Thank you again.

PR #90, opened 2026-07-13
fix(opencode): use platform global data dir instead of project dir

cawa0505:
Move token-optimizer data storage from project-local to platform global locations.

**Problem:** token-optimizer stored session DBs and trends in `{project}/token-optimizer/`, which polluted git and prevented cross-project trend aggregation.

**Solution:** 
- Data now lives at `~/.local/share/token-optimizer/` (Linux), `~/Library/Application Support/token-optimizer/` (macOS), or `%LOCALAPPDATA%/token-optimizer/` (Windows)
- Overridable via `TOKEN_OPTIMIZER_DATA_DIR` env var or `PluginOptions.dataDir`
- Session DBs scoped per project under `sessions/{project-slug}/` using Claude Code's path-encoding convention (`/Users/alex/my-project` → `-Users-alex-my-project`)
- Existing paths inside storage classes unchanged — just the base `dataDir` resolution changed

**Files changed:**
- `util/env.ts` — added `resolveDataDir()` + `hashProjectDir()`
- `index.ts` — use `resolveDataDir()` instead of `ctx.directory`
- `storage/session-store.ts` — accept `projectSlug` for scoped session dirs
- `continuity/restore.ts` — accept `projectSlug`, scope scan
- `continuity/resume-lean.ts` — accept `projectSlug`, scope scan

```
~/.local/share/token-optimizer/
├── trends.db
├── sessions/
│   ├── -Users-alex-my-project/
│   │   └── {sessionId}.db
│   └── ...
└── dashboard.html
```

### DIFF
--- a/opencode/bun.lock
+++ b/opencode/bun.lock
@@ -1,5 +1,6 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "token-optimizer-opencode",
--- a/opencode/src/continuity/restore.ts
+++ b/opencode/src/continuity/restore.ts
@@ -16,6 +16,9 @@ export function restoreCheckpoint(
   userPrompt: string,
   currentSessionId: string,
   config: TokenOptimizerConfig,
+  /** Project slug for scoping session scans under sessions/{slug}/. Derived
+   *  from ctx.project.worktree by the caller (hashProjectDir). */
+  projectSlug?: string,
   /** Optional: when supplied and the prompt shows resume-intent, the lean
    *  reconstruction path fires and savings are credited to trends.db. */
   trendsStore?: TrendsStore,
@@ -25,7 +28,9 @@ export function restoreCheckpoint(
 ): CheckpointMatch | null {
   if (!config.features.continuity) return null;
 
-  const sessDir = join(dataDir, "token-optimizer", "sessions");
+  const sessDir = projectSlug
+    ? join(dataDir, "token-optimizer", "sessions", projectSlug)
+    : join(dataDir, "token-optimizer", "sessions");
   if (!existsSync(sessDir)) return null;
 
   // DB filenames are the sanitized session id, so compare like-for-like or a
@@ -44,6 +49,7 @@ export function restoreCheckpoint(
         userPrompt,
         dataDir,
         safeCurrentId,
+        projectSlug,
         cwd,
         config.checkpointRetentionDays,
         config.checkpointRetentionMax,
--- a/opencode/src/continuity/resume-lean.ts
+++ b/opencode/src/continuity/resume-lean.ts
@@ -410,13 +410,17 @@ export function buildResumeLeanBlock(
   userPrompt: string,
   dataDir: string,
   currentSessionId: string,
+  /** Project slug for scoping session scans under sessions/{slug}/. */
+  projectSlug: string | undefined,
   cwd: string,
   retentionDays: number = 7,
   maxCandidates: number = 50,
 ): [string, string] {
   if (!cwd) return ["", ""];
 
-  const sessDir = join(dataDir, "token-optimizer", "sessions");
+  const sessDir = projectSlug
+    ? join(dataDir, "token-optimizer", "sessions", projectSlug)
+    : join(dataDir, "token-optimizer", "sessions");
   if (!existsSync(sessDir)) return ["", ""];
 
   const candidates = loadSameProjectCheckpoints(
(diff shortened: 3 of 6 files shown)


alexgreensh:
Thanks for this PR! The intent (moving to platform-global storage) is sound and the security posture is solid — `hashProjectDir` correctly matches Claude Code's encoding convention, and the per-project session scoping prevents cross-project data leakage.

However, there are blocking issues that need to be addressed before merge:

## Critical

**C1. Signature changes break existing tests.** `restoreCheckpoint` and `buildResumeLeanBlock` insert `projectSlug` as the 5th param, shifting `trendsStore` and `cwd`. Existing test calls still use the old order, so `cwd` (a string) lands in the `trendsStore` slot and the actual `cwd` param is `undefined` — the resume-lean path never fires. The PR includes zero test file changes.

**C2. `resolveDataDir` ignores `ctx.directory`.** Integration tests pass a temp dir via `mockCtx(dir)` expecting data to land there, but `resolveDataDir(config)` returns the global platform path. All integration tests that assert on persisted data fail. Fix: wire `config.dataDir` in `resolveConfig` from `opts.dataDir`, or have tests set `TOKEN_OPTIMIZER_DATA_DIR`.

## High

**H1. `config.dataDir` is dead code.** The interface adds `dataDir?: string` and `resolveDataDir` checks it, but `resolveConfig()` never populates it from `opts.dataDir`. The only override path is the env var. Wire it in `resolveConfig` (this also fixes C2) or remove it from the interface.

**H2. No migration path.** Existing users have `trends.db` and session DBs in `{project}/token-optimizer/`. After this PR, data is read/written from the global location with no migration, no fallback read, no warning. Users silently lose their trends history and continuity checkpoints. At minimum, first-run should check the old location and migrate or log a warning.

## Medium

**M1. Trailing-slash stripping is too aggressive.** The regex `/\/?token-optimizer\/?$/` matches `bar-token-optimizer` (strips to `bar-`) because the leading separator is optional. Fix: require a path separator boundary — `/[\/\\]token-optimizer[\/\\]?$/`.

**M2. No README update** for the storage location change or `TOKEN_OPTIMIZER_DATA_DIR` override.

## Low

- **L1:** `ctx.project.worktree ?? process.cwd()` — per SDK types, `worktree` is `string` (required), so the fallback is dead code. Consider using `ctx.worktree` (top-level `PluginInput` field) instead.
- **L2:** `bun.lock` `configVersion: 0` — harmless bun version artifact.
- **L3:** "ponytail" comment → should be `TODO:`.
- **L4:** No unit tests for `resolveDataDir` or `hashProjectDir`.

## CLA

The CLA check is still failing — that needs to be signed before we can merge.

---

**Summary of required changes before merge:**
1. Fix test breakage (C1+C2) — wire `config.dataDir` in `resolveConfig` and update test calls
2. Add migration or warning (H2)
3. Fix trailing-slash regex (M1)
4. Wire or remove `config.dataDir` (H1)
5. Add unit tests (L4)
6. Sign CLA

alexgreensh:
@cawa0505 — I owe you a correction before anything else.

Two of the items I labelled "Critical" in my earlier review do not exist. I said your signature changes broke existing tests, and that integration tests using `mockCtx` would fail against `resolveDataDir`. Neither is true. This package has exactly one pre-existing test file (`savings.baseline.test.ts`, covering `computeRealizedSavings`), `mockCtx` appears nowhere in the repository, and nothing outside `src/` calls `restoreCheckpoint` or `buildResumeLeanBlock`. That review was wrong, it was wrong in the two places I marked most severe, and it sat on your PR for days. Sorry.

What was accurate: the trailing-slash regex (M1), `config.dataDir` never being wired in `resolveConfig` (H1), and the missing migration path (H2).

On the outcome: this shipped in **v5.11.51** (opencode 1.1.0) as an independent implementation rather than a merge of yours, because the CLA was never signed and I could not merge unsigned code into a dual-licensed codebase. To be explicit, I did not copy your diff. I implemented from the problem statement, and several decisions came out differently:

- Optional parameters are appended rather than inserted mid-signature, so no existing call site shifts its arguments.
- Suffix stripping is segment-aware instead of regex-based, so `my-token-optimizer` cannot be truncated to `my-`.
- Migration copies rather than moves, leaving the originals recoverable.

A QA pass also surfaced three things worth naming, since they were latent in the shared design rather than in your patch specifically:

- Migration had to write session DBs into `sessions/{slug}/`, not flat into `sessions/`, or every migrated session would be invisible to the scoped readers while the log reported success.
- WAL sidecars must be copied alongside each `.db`. Both stores run `PRAGMA journal_mode=WAL`, so uncheckpointed rows live in `-wal`.
- All migration filesystem checks use `lstat`. Following symlinks would let a hostile repo ship `token-optimizer/trends.db -> ../../../../.ssh/id_ed25519` and have it copied into the data directory on plugin init.

The diagnosis and the direction are yours. Global platform dirs, per-project session scoping, and matching the path-encoding convention were all your calls and they were right. You are credited in the commit trailer, the release notes, and the test file that pins the behaviour.

If you would like to keep contributing, signing the CLA once unblocks everything future, and I will give the next one a review that has actually read the test suite.

Closing this in favour of the shipped implementation. Thanks for the report.

PR #96, opened 2026-07-22
fix: sync TOKEN_OPTIMIZER_VERSION constant to 5.11.53

danikdanik:
Hey! Small one-line-times-two fix I noticed while poking at the dashboard today.

**What**
`TOKEN_OPTIMIZER_VERSION` inside `measure.py` (both copies — `skills/token-optimizer/scripts/measure.py` and `plugins/token-optimizer/skills/token-optimizer/scripts/measure.py`) is hardcoded to `"5.11.50"`, while `.claude-plugin/plugin.json` and `.claude-plugin/marketplace.json` are both already at `"5.11.53"`. This PR just brings the constant back in sync — no behavior change, no other lines touched.

**Why**
That constant drives the dashboard header (`Token Overhead Analysis vX.X.X`), so right now a fresh 5.11.53 install still shows `v5.11.50` in the UI. Confirmed on a clean install today: `installed_plugins.json` said 5.11.53, the dashboard said 5.11.50. Also looks like a recurring release-process gap, not a one-off — I checked an older fork checkout at 5.11.37 and the constant was similarly behind (`5.11.35`) there too, so this has drifted across at least two releases. Might be worth a follow-up at some point to derive this from `plugin.json` instead of hand-maintaining a third copy (the comment on that line already says "Keep in sync with plugin.json + marketplace.json" — this makes it true again, at least for now).

**How**
Two-line diff, straight string swap, `"5.11.50"` → `"5.11.53"` in both copies. Nothing else in either file changed. Verified with `git diff` before pushing.

Thanks for building this, it's genuinely useful — figured I'd send the fix rather than just file an issue since it was this small. Happy to open a follow-up issue for the "derive it from plugin.json" idea if that's useful, or drop it if you'd rather keep it simple.

### DIFF
--- a/plugins/token-optimizer/skills/token-optimizer/scripts/measure.py
+++ b/plugins/token-optimizer/skills/token-optimizer/scripts/measure.py
@@ -18552,7 +18552,7 @@ def setup_hook(dry_run=False, uninstall=False):
 
 # ========== Persistent Dashboard Daemon ==========
 
-TOKEN_OPTIMIZER_VERSION = "5.11.50"  # Keep in sync with plugin.json + marketplace.json
+TOKEN_OPTIMIZER_VERSION = "5.11.53"  # Keep in sync with plugin.json + marketplace.json
 _DASHBOARD_CSP = "default-src 'none'; script-src 'unsafe-inline'; style-src 'unsafe-inline' https://fonts.googleapis.com; font-src https://fonts.gstatic.com; connect-src 'self'; img-src 'self' data:; base-uri 'none'; form-action 'none'; frame-ancestors 'none'"
 # Per-runtime daemon identity. Each runtime gets a distinct port + label so a
 # dashboard under one runtime never collides with another's. Copilot uses 24845
--- a/skills/token-optimizer/scripts/measure.py
+++ b/skills/token-optimizer/scripts/measure.py
@@ -18552,7 +18552,7 @@ def setup_hook(dry_run=False, uninstall=False):
 
 # ========== Persistent Dashboard Daemon ==========
 
-TOKEN_OPTIMIZER_VERSION = "5.11.50"  # Keep in sync with plugin.json + marketplace.json
+TOKEN_OPTIMIZER_VERSION = "5.11.53"  # Keep in sync with plugin.json + marketplace.json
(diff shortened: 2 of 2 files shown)


danikdanik:
@alexgreensh , one more thought on the "keep in sync" comment on that line, in case it's useful: since this has now drifted twice across releases (this PR + an older 5.11.37 install where the constant was similarly behind), a one-line fix each time probably won't stick on its own. Two options if you want to close the door on it for good, cheapest first:

1. **CI check** — a small workflow step that fails if `TOKEN_OPTIMIZER_VERSION` in either `measure.py` copy doesn't match `.claude-plugin/plugin.json`'s `version`. Catches it at PR time, no runtime code touched.
2. **Derive instead of duplicate** — have `measure.py` read the version from `plugin.json` at import time instead of carrying a third hardcoded copy. Removes the failure mode entirely, but touches live code in both `measure.py` locations, so more to review.

No pressure either way, just flagging it since I stumbled into the same class of bug twice while poking around. Happy to send a follow-up PR for #1 if that'd save you the time.

alexgreensh:
Thanks Dani, this was a better catch than a two-line diff suggests.

You were right that it was a release-process gap rather than a one-off. By the time this landed the constant was four releases behind (`5.11.50` against manifests at `5.11.54`), which matches the drift you spotted in the 5.11.37 checkout.

Rather than merge the literal bump, I took the follow-up you suggested in the description: `TOKEN_OPTIMIZER_VERSION` now derives from `.claude-plugin/plugin.json` instead of being a third hand-maintained copy. The line carried a "keep in sync with plugin.json + marketplace.json" comment that had not been true for four releases, and any value a human has to remember to update on release day will eventually be missed. Lookup walks up to the nearest manifest and falls back to a default for skill-only installs, and never raises.

It proved itself immediately: bumping the manifests to 5.11.55 propagated to the dashboard header with no second edit.

The part worth flagging back to you: the repo's own version audit should have caught this and did not. It truncates files at 500,000 characters as a speed guard, and `measure.py` is 1.6MB with the constant at line 18,645. The scanner read the first 30% of the file, found no version, and reported OK. So the gate did not miss the drift, it never looked, and then reported everything fine. That is fixed too, plus a test that fails if the constant ever becomes a literal again.

Shipped in v5.11.55 and credited in the release notes. Closing this in favour of the generalized fix, but the find is yours. If you want to open that follow-up issue for anything else in the release path that looks hand-maintained, please do.

PR #100, opened 2026-07-26
fix: report CLAUDE.md offenders per-file instead of summing across ancestors

danikdanik:
## Problem

The "Slim CLAUDE.md" quick win can report a token/line count that isn't any single file's size. On my own dashboard it showed "12,058 tokens, target ~4,500 / ~300 lines" as if one file were bloated. It wasn't one file — it was three separately-owned CLAUDE.md files summed together: my global ~/.claude/CLAUDE.md (1,254 tok), a workspace-root CLAUDE.md (1,149 tok), and a project's own CLAUDE.md pulled in because measure_components() walks cwd.parents[:3] (9,655 tok). 1254+1149+9655 = 12,058 exactly, confirmed against dashboard.html's window.__TOKEN_DATA__.

## Root cause

measure_components() (measure.py:1455-1476) already keys each ancestor file with its own identity: claude_md_project_{parent.name}. The data model is fine.

The bug is downstream: every consumer throws that identity away by summing indiscriminately over every claude_md*-prefixed key instead of using it:

- generate_auto_recommendations (6172-6198) — the quick win itself
- quick_scan (2601-2608, 2669-2676) — "top offenders" + the terse action text
- _auto_snapshot (2545-2548) — trend history
- claude_md_health (5684-5705) — the dashboard health card

Any monorepo/worktree layout with 2-3 CLAUDE.md files up the tree produces a combined number and a "slim CLAUDE.md to ~300 lines" recommendation that isn't actionable on any one file — you can't edit three files someone else may own down to a shared line target.

## What this changes

- generate_auto_recommendations and quick_scan's offender entry now report per-file: when a single claude_md_* component exceeds the threshold, the recommendation names that specific file and its own line count, instead of blending N files under one "Slim CLAUDE.md" label.
- No change to measure_components()'s ancestor walk itself — that mirrors how Claude Code actually loads CLAUDE.md files, so collecting per-ancestor data is correct. Only the summarization on top of it was wrong.

## Note

@alexgreensh , since this changes recommendation behavior/UX, not just a factual correction (that's PR #99). Flag if you'd rather go a different direction - e.g. only ever treating cwd's own file + global as "yours to slim," dropping ancestor files from the recommendation entirely instead of listing them per-file

### DIFF
--- a/plugins/token-optimizer/skills/token-optimizer/scripts/measure.py
+++ b/plugins/token-optimizer/skills/token-optimizer/scripts/measure.py
@@ -2598,17 +2598,18 @@ def quick_scan(as_json=False):
         if eager > 0:
             detail += f" ({eager} with eager-loaded tools)"
         offenders.append(("mcp", mcp_count, mcp.get("tokens", 0), detail))
-    claude_md_tokens = sum(
-        components[k].get("tokens", 0)
-        for k in components if k.startswith("claude_md") and components[k].get("exists")
-    )
-    claude_md_lines = sum(
-        components[k].get("lines", 0)
-        for k in components if k.startswith("claude_md") and components[k].get("exists")
-    )
-    if claude_md_tokens > 0:
-        offenders.append(("claude_md", claude_md_lines, claude_md_tokens,
-                         f"CLAUDE.md ({claude_md_lines} lines)"))
+    # Per-file CLAUDE.md offenders (don't blend multiple ancestor files into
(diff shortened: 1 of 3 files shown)


alexgreensh:
Thank you @danikdanik — this was the right call. Your per-file CLAUDE.md offender reporting shipped in **v5.11.69**, reconciled onto the current tree and credited via `Co-authored-by` on the commit. Closing since it's now released. Much appreciated.

PR #102, opened 2026-07-28
Add current git branch to status line

ofergr:
## Summary
- Adds the current git branch name to the status line, shown alongside the project directory.

## Test plan
- [x] Ran `node --check` on statusline.js
- [x] Piped a sample stdin payload through the script from inside a git repo and confirmed the branch segment renders correctly
- [x] Confirmed no branch segment appears when run outside a git repo

### DIFF
--- a/skills/token-optimizer/scripts/statusline.js
+++ b/skills/token-optimizer/scripts/statusline.js
@@ -13,6 +13,7 @@
 const fs = require('fs');
 const path = require('path');
 const os = require('os');
+const { execFileSync } = require('child_process');
 
 let input = '';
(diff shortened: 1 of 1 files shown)


ofergr:

ofergr:
recheck

alexgreensh:
Shipped in **v5.11.68** — thank you @ofergr! Your commit is in with authorship preserved. One addition on top: `windowsHide: true` on the git spawn, so it doesn't flash a console window on Windows (we shipped a fix for that exact class in v5.11.67). The branch now shows in the status line right after the project name. 🙏

PR #125, opened 2026-08-08
fix: gate quality-cache --warn with the same cap/cooldown/band guardrails as _maybe_nudge

### LINKED ISSUE #124: quality-cache --warn re-nags on every prompt: no cap/cooldown/band-drop gate (unlike the sibling nudge)
## Summary

`quality-cache --warn` — the plain-text `[Token Optimizer] Context quality: N/100...` message installed on the `UserPromptSubmit` hook (`quality-cache --warn --quiet`) — has no session cap, no cooldown, and no "only if it got worse" check. It re-prints on **every single prompt** for as long as the score stays below `warn_threshold` (default 70).

## Repro

1. Get a session's quality score into, say, the low-to-mid 60s (any mix of stale reads / bloated tool results / low signal ratio works).
2. Keep working normally — score wobbles around in that band, e.g. `69 -> 66 -> 64 -> 62`, never crossing back above 70.
3. Every `UserPromptSubmit` re-triggers `measure.py quality-cache --warn --quiet`, and the CLI unconditionally re-prints:

   ```
   [Token Optimizer] Context quality: 66/100. Stale reads and bloated results building up. Consider /compact.
   ```

   ...on every prompt, even though nothing meaningfully changed since the last warning. Over a long session this is a constant nag.

## Root cause

In `skills/token-optimizer/scripts/measure.py`, the `--warn` emit was gated only by:

```python
if warn and score is not None and score < warn_threshold:
    ...
    if not should_inject(session_jsonl, priority="informational"):
        ...
    else:
        print(f"[Token Optimizer] Context quality: {score}/100...")
```

`should_inject` (context-pressure throttle) is a belt-and-suspenders check, not a dedicated cap/cooldown for this message — it doesn't track this warning's own fire history at all.

Compare this to the sibling proactive nudge, `_maybe_nudge` (a few hundred lines above the `--warn` CLI dispatch), which fires the JSON `systemMessage` "Quality dropped to N... Consider /compact to protect context." nudge. That path already has:

- `_NUDGE_SESSION_CAP` (3 fires per session)
- `_NUDGE_COOLDOWN_SECONDS` (300s between fires)
- Only fires on a >15pt drop, or on crossing below 60

So the author's clear intent is capped, non-repeating nudges — the `--warn` CLI message is the one path that never got the same guardrails. The two nudge paths are inconsistent with each other for no apparent reason.

## Proposed fix

Give `--warn` the same guardrails as `_maybe_nudge`, reusing `_NUDGE_SESSION_CAP` / `_NUDGE_COOLDOWN_SECONDS`, plus a band-drop gate so it only re-fires when the score crosses into a **new, lower** 10-point band:

```python
_band = int(score // 10)
if count >= _NUDGE_SESSION_CAP or now - last_epoch < _NUDGE_COOLDOWN_SECONDS or _band >= last_band:
    _emit_warn = False
# on emit: persist last_epoch=now, count+1, last_band=_band
```

So `69 -> 66 -> 64 -> 62` (all band 6) warns once, and a later drop into the 50s (band 5) warns again. State needs to persist per-session across the separate hook processes (the existing quality-cache sidecar / `_CARRY_KEYS` mechanism already does this for `_maybe_nudge`'s own state).

I have a PR ready that implements this.

eligrumman:
## Summary

`quality-cache --warn` — the plain-text warning the `UserPromptSubmit` hook installs (`quality-cache --warn --quiet`) — had no session cap, no cooldown, and no "only if it got worse" check. It fired on every prompt where `score < warn_threshold` (default 70), so a session hovering in a single 10-point band (e.g. `69 -> 66 -> 64 -> 62`) nagged the user on every single turn. Fixes #124.

Its sibling, `_maybe_nudge` (the proactive `systemMessage` nudge a few hundred lines earlier in `measure.py`), already solved this exact problem: it enforces `_NUDGE_SESSION_CAP` (3) and `_NUDGE_COOLDOWN_SECONDS` (300s), only firing on a >15pt drop or a cross below 60. The `--warn` path never got the same guardrails — the two nudge paths were inconsistent with no apparent reason for the gap.

## What changed

- Added `_maybe_quality_warn(result, warn_threshold)` next to `_maybe_nudge`, gated the same way: reuses `_NUDGE_SESSION_CAP` / `_NUDGE_COOLDOWN_SECONDS`, plus a band-drop check — it only re-fires once the score crosses into a **new, lower** 10-point band. `69 -> 66 -> 64 -> 62` now warns once; a later drop into the 50s warns again.
- State (`_warn_count` / `_warn_last_epoch` / `_warn_last_band`) persists in the existing per-session quality-cache sidecar, registered in `_CARRY_KEYS` alongside the other nudge/loop/warning state — the same mechanism `_maybe_nudge` already relies on to survive the `UserPromptSubmit` → `PostCompact` process boundary.
- The actual `print(...)` moved from the CLI dispatch (`elif args[0] == "quality-cache":`) into `quality_cache()` itself, gated behind a new `warn=False` parameter. `quality_cache()` is only called from that one CLI dispatch site, so this is a safe, non-breaking move — and it's actually *more* consistent with the file's existing style: every other nudge/loop/fill/tool-call warning already prints from inside `quality_cache()`, where `filepath`/`cache_path`/`result` are already resolved; the plain-text `--warn` message was the one outlier printing from outside it, re-deriving nothing extra but duplicating the `should_inject` gate pattern one level removed.
- The existing `context_pressure.should_inject` throttle is kept exactly as-is, still applied unconditionally as a second, independent gate (belt-and-suspenders).
- `plugins/token-optimizer/skills/token-optimizer/scripts/measure.py` (the Codex marketplace mirror) re-synced via `scripts/sync-codex-marketplace-plugin.sh` so the two trees stay byte-identical (enforced by `tests/test_duplicate_script_sync.py` / `tests/test_session_locale_and_hooks.py::test_measure_py_dual_tree_parity`).

## Why this is low-risk

- No default thresholds changed, no behavior removed — the warning still fires, just not on every single prompt.
- `warn` defaults to `False`, so `quality_cache()`'s other caller pattern (`--force --quiet` from `SessionStart`/`PreCompact`, which never passes `--warn`) is byte-for-byte unaffected.
- `score` is not otherwise used at the CLI dispatch call site after this change, and printing already happens exactly where every other nudge in this file prints from.

## Testing

- Added `tests/test_quality_warn_band_gate.py`:
  - Unit tests directly against `_maybe_quality_warn` (first-drop fires, same-band repeats are suppressed, a genuine drop into a new lower band re-fires, cooldown blocks even a real band drop, session cap stops firing after 3, at/above threshold never fires).
  - Two end-to-end tests driving the real `quality_cache(warn=True)` path (mirroring `tests/test_quality_cache_release_on_timeout.py`'s pattern) with `capsys`, proving the actual CLI print fires once across three same-band ticks and fires again on a real band drop.
  - A sanity test that `warn=False` never prints.
- `python3 -m pytest tests/ -q` (excluding one file/test that fails identically on unmodified `main` under this Python 3.9 environment — `dict | None` syntax needs 3.10+, unrelated to this change): **1052 passed, 13 skipped**.
- `python3 -m py_compile` on both copies of `measure.py` and the new test file: clean.

Fixes #124

### DIFF
--- a/plugins/token-optimizer/skills/token-optimizer/scripts/measure.py
+++ b/plugins/token-optimizer/skills/token-optimizer/scripts/measure.py
@@ -30886,9 +30886,10 @@ _LOOP_LAST_MESSAGES = 4
 # Transient nudge/loop/warning state that must survive across the separate
 # UserPromptSubmit and PostCompact hook processes. compute_quality_score() builds
 # a fresh result each run, so quality_cache() carries these forward from the prior
-# on-disk cache. Any new "_nudge_*"/"_loop_*"/warning-dedup key set in _maybe_nudge
-# or _maybe_loop_warning MUST be registered here — a sibling key silently missing
-# from this set is exactly what broke the nudge follow-through credit (A6).
+# on-disk cache. Any new "_nudge_*"/"_loop_*"/"_warn_*"/warning-dedup key set in
+# _maybe_nudge, _maybe_loop_warning, or _maybe_quality_warn MUST be registered
+# here — a sibling key silently missing from this set is exactly what broke the
+# nudge follow-through credit (A6).
 _CARRY_KEYS = (
     "_nudge_fill_pct_at_fire",
     "_nudge_fire_epoch",
@@ -30900,6 +30901,9 @@ _CARRY_KEYS = (
     "progressive_bands_captured",
     "_last_fill_warn_level",
     "_last_tool_call_warn_level",
+    "_warn_count",
+    "_warn_last_epoch",
+    "_warn_last_band",
 )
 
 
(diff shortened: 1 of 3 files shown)


eligrumman:
Superseded by #126 (same fix, fresh branch `fix/quality-warn-cap-cooldown-band`). Closing this one.

PR #132, opened 2026-08-10
fix: #129 cold-resume safety + dedupe savings-card copy

### LINKED ISSUE #129: prompt-continuity cold-resume-lean can inject an unrelated session's checkpoint with an unconditional "tell the user you reopened it" instruction
## Summary

In a workspace with multiple parallel Claude Code sessions open on the same project directory, the `prompt-continuity` UserPromptSubmit hook (cold-resume-lean / "keep-warm v2") can inject a **different, unrelated session's** checkpoint content into the current session, together with an unconditional instruction telling the assistant to claim to the user that it "reopened" that session. Observed in practice: the injected checkpoint belonged to a sibling session that was simply the most recently active one in the same project directory at that moment -- not the session the user meant, and not the session the user's own prompt explicitly named by id.

## Root cause

`_continuity_resume_block()` (measure.py, ~line 24766) triggers when the prompt matches `_RESUME_INTENT_RE` (phrases like "continue our/the/this/that/from...", "last session", "we were working on", etc. -- a broad, easy-to-hit set). It:

1. Filters checkpoints to the same project directory (cwd-prefix match) -- never by conversation/session identity.
2. Scores same-project checkpoints by keyword/topic overlap against the prompt text.
3. **When nothing clears the relevance bar (`_RESUME_NAMED_TOPIC_BAR`, default 0.22) -- i.e. a vague "continue" with no distinguishing topic -- it silently falls back to `max(same_project, key=lambda x: x[1]["created"].timestamp())`: whichever sibling session in the same project directory has the freshest checkpoint, with no relevance floor and no staleness cap** (`_best_checkpoint_for_session` is called with the default `within_retention=False`, so even a very old checkpoint qualifies).
4. Nowhere in the file is an explicit session id parsed out of the prompt text, even though a very natural way to ask for resume ("continue the work of session `<id>`") includes one. If the user literally names the session they mean, that id is never consulted -- the vague-ask fallback can still silently substitute a different session.
5. The resulting block (`build_lean_resume_context`, ~line 25189) ends with an **unconditional** footer: "Use this to re-orient a fresh session on the prior work. Tell the user you reopened the cold session (mention its date/topic) so the recovery is transparent." This asserts the match is correct rather than asking the assistant to verify it first.

By contrast, the SessionStart `compact-restore --new-session-only` path (`compact_restore()`, ~line 24363), which does similar cross-session lookup, is safe by construction: it enforces a 30-minute freshness cutoff, emits only a one-line pointer + short descriptor (never the checkpoint body), and phrases it conditionally: "Load it only if it matches what you are working on now." The plugin's own *lightweight hint* branch (same file, ~line 24955-25023, used when no resume-intent phrase is detected) also already uses safe, conditional wording ("...tell the user you found a relevant prior session... so the recovery is transparent, not silent"). The bug is specifically that the "deliberate continue" full-reconstruction branch follows neither of those two already-established safe patterns elsewhere in the same file.

## Suggested fix (happy to open a PR if useful)

1. Replace the unconditional footer at ~line 25282-25286 with the same conditional wording already used in the lightweight-hint branch.
2. When no checkpoint clears `_RESUME_NAMED_TOPIC_BAR`, don't fall back to blind recency across an arbitrary sibling session -- either return `""` (there's already a "no same-project match -> return nothing" path a few lines below for a similar case) or require the fallback to pass the same freshness gate `compact_restore` already enforces, downgraded to the lightweight-hint payload/wording.
3. Parse an id-shaped token (Claude Code session ids are UUIDs, often referenced by an 8-char prefix) out of the prompt text and, if present, scope strictly to that session -- and say so explicitly if it matches no on-disk checkpoint, rather than silently substituting an unrelated one.

## Environment

- Plugin version: 5.11.15
- Trigger condition: several Claude Code sessions open concurrently against the same project directory (this is what exposes the bug -- the "most recent in project" fallback degrades to "most recently active sibling session" rather than anything related to conversational continuity).

alexgreensh:
## What

Two safe, reviewed fixes:

### 1. Fix #129 — cold-resume no longer injects an unrelated session's checkpoint
The `prompt-continuity` UserPromptSubmit hook could reconstruct a **different** session's checkpoint (whichever sibling was last active in the same project dir) and tell the assistant, unconditionally, to claim it "reopened" that session. Reproduced live: a fresh session opened with an injected checkpoint from an unrelated `/reload-plugins` cleanup session.

Three changes in `measure.py`:
- **Conditional footer.** `build_lean_resume_context` gains `footer_mode`. The auto-inject path (a *guess*) now uses verify-before-claiming wording mirroring the lightweight-hint/`compact_restore` branches; the CLI `--resume-lean` and explicit-id paths stay confident.
- **Staleness-capped recency.** The vague-"continue" fallback only surfaces the freshest same-project checkpoint when it is newer than `TOKEN_OPTIMIZER_RESUME_RECENCY_MIN` (default 480 min); otherwise it declines (`""`) rather than blindly reopening a stale sibling.
- **Explicit session id honored.** A session id named in the prompt ("continue session `<id>`", full UUID) is scoped strictly to that session and never substituted; the resume-intent gate now also fires on a named id (it previously rejected the exact phrasing), and the match resolves by filename (not a possibly-disagreeing sidecar id).

### 2. Dashboard: dedupe the methodology wall-of-text
The savings-card footnote rendered the "Per-window dollars reuse the metered savings ledger…" sentence **twice** (once from `rw.proxy`, once hardcoded in the HTML) plus a redundant "usage is measured" prefix. Consolidated into a single authoritative note; the HTML now adds only the one live-value fact it must.

## Tests
- New `tests/test_129_cold_resume_safety.py`: 11 tests covering footer routing, the staleness cap, explicit-id scoping, and (per code review) the **real hook path** for "continue session `<id>`", incidental-id fall-through, and filename-vs-sidecar resolution.
- Full suite: **1188 passed, 13 skipped**.

## Review
- Sol (Codex, cross-model adversarial): no findings.
- CE correctness lens: caught that the explicit-id branch was gated behind a resume-intent regex that rejected its own phrasings (fixed; regression-tested through the real hook entry).
- Not in scope (separate change per plan): dashboard auto-regen, stuck-session hook latency.

### DIFF
--- a/plugins/token-optimizer/skills/token-optimizer/assets/dashboard.html
+++ b/plugins/token-optimizer/skills/token-optimizer/assets/dashboard.html
@@ -3953,8 +3953,13 @@ var DB = (function() {
             '<div id="to-regen-error" hidden style="margin-top:10px;padding:10px 14px;font-size:15px;font-weight:500;color:var(--c-waste);background:rgba(255,95,95,.08);border:1px solid rgba(255,95,95,.35);border-radius:8px;line-height:1.5;text-wrap:pretty;"></div></div>';
         })() +
         '<div style="margin-top:var(--s-4);padding-top:var(--s-3);border-top:1px solid var(--c-border);font-size:14px;color:var(--c-text-dim);line-height:1.7;text-wrap:pretty;">' +
-          'Your usage is measured. The window comparison and the routing dollars are estimated: ' + esc(rw.proxy) + ' ' +
-          'Per-window dollars reuse the metered savings ledger (context tokens never sent, priced at input rates) plus an estimated model-routing counterfactual; they are not derived from the throughput multipliers. The &asymp;' + rw.extra_work_pct + '% is a separate per-token throughput multiplier, not a sum of the per-window figures.' +
+          // rw.proxy is now the single source of the measured-vs-estimated boundary
+          // AND the ledger detail (deduped: this used to repeat the ledger sentence
+          // and re-state "usage is measured"). We add only the one fact proxy can't
+          // carry -- the throughput-% is its own unit, not a sum of the dollars --
+          // because it needs the live JS value.
+          esc(rw.proxy) + ' ' +
+          'The &asymp;' + rw.extra_work_pct + '% is a separate per-token throughput multiplier, not a sum of the per-window dollar figures.' +
         '</div>' +
       '</div>';
     }
--- a/plugins/token-optimizer/skills/token-optimizer/scripts/measure.py
+++ b/plugins/token-optimizer/skills/token-optimizer/scripts/measure.py
@@ -28664,6 +28664,39 @@ _RESUME_INTENT_RE = re.compile(
 # session") so we fall back to the most-recent same-project checkpoint.
 _RESUME_NAMED_TOPIC_BAR = _float_env("TOKEN_OPTIMIZER_RESUME_TOPIC_BAR", 0.22)
 
+# Staleness cap for the VAGUE-continue fallback (GitHub #129): when the prompt
+# names no distinguishing topic we may only surface the most-recent same-project
+# checkpoint if it is fresher than this. Beyond it, blindly reopening whichever
+# sibling session was last active in the same folder is how an UNRELATED session's
+# checkpoint gets injected -- so we return nothing instead of guessing. The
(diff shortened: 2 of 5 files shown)


alexgreensh:

PR #135, opened 2026-08-10
fix(dashboard): live-refresh the 5h/7d window card (auto-regen)

alexgreensh:
**Problem:** the savings card's window figure went stale and the Regenerate button couldn't refresh it — the card even says so. The 5h/7d reading is only fresh while the statusline writes it (active turns), and the dashboard's existing live path (`GET /api/savings`) patched every tile *except* the window card.

**Fix:** add `runway` to the live `_live_savings_payload` (guarded, fail-open), patch it in `refreshSavings`, and poll every ~60s (skipped when the tab is hidden). An open dashboard now reflects a new reading on its own — no manual Regenerate.

**Honesty boundary (kept):** this doesn't invent a number while Claude is idle (the windows exist only in the statusline stdin); it surfaces the newest *real* reading automatically, and the "as of / N ago" label keeps counting truthfully.

**Ops:** the installed daemon (v5.11.85) predates `/api/savings`, so it needs a regenerate/restart for the live path to work on an existing machine (separate step, called out in the commit).

**Tests:** `tests/test_dashboard_live_runway.py` — runway in payload (== builder), fail-open, live-reflect, client wiring mirrored to plugin tree. Full suite 1184 passed.

### DIFF
--- a/plugins/token-optimizer/skills/token-optimizer/assets/dashboard.html
+++ b/plugins/token-optimizer/skills/token-optimizer/assets/dashboard.html
@@ -6255,6 +6255,13 @@ var DB = (function() {
     if (payload.cache_health && typeof payload.cache_health === 'object') {
       data.cache_health = payload.cache_health;
     }
+    // Live window/runway card: pick up a fresher 5h/7d reading the instant the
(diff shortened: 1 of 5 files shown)


alexgreensh:

PR #136, opened 2026-08-10
fix(hooks): budget the compact-restore SessionStart hook (first-prompt stall)

alexgreensh:
**Problem:** a brand-new session could hang ~36s on the first prompt when several Claude Code sessions were open on the same project dir.

**Root cause (measured):** not the continuity scan (that's checkpoint-bounded and 8s-budgeted). It's the synchronous first-prompt hook chain, dominated by `compact-restore --new-session-only` — the **only** hook with no internal budget, capped solely by the harness's 20s timeout (2.5x the 8s cap on every other hook). Under disk/CPU contention it dominated the stall.

**Fix:** wrap it in the same `_install_hook_budget(...)` guard the other eight hooks already use (env-overridable, default 8s). Its output is a best-effort re-orientation hint, so a budget-exit degrades to no hint instead of blocking.

**Deliberately deferred** (bigger/riskier): collapsing the 3 UserPromptSubmit hooks into one process, caching the O(all-jsonl) session-resolution fallback (a shared cross-process cache risks misattributing the current session in exactly the parallel-session case that triggers this), and globally lowering budgets. This PR caps the single dominant contributor with a low-risk, pattern-matched edit; the additive-chain worst case is a follow-up worth doing deliberately.

**Tests:** clean exit under default + tiny budget, edit-time budget guard (both source trees). Full suite 1181 passed.

Honest caveat: I could not reproduce the full 36s live (warm, every hook 0.35–0.70s) — it's the cold/contended tail these ceilings allowed. This removes the largest ceiling.

### DIFF
--- a/plugins/token-optimizer/skills/token-optimizer/scripts/measure.py
+++ b/plugins/token-optimizer/skills/token-optimizer/scripts/measure.py
@@ -39315,19 +39315,32 @@ if __name__ == "__main__":
             else:
                 compact_restore(session_id=sid, is_compact=is_compact)
 
-        # Codex requires SessionStart stdout to be empty or valid JSON. Under the
-        # Codex marketplace plugin the shared hooks.json calls this directly (not
-        # via codex_hook_bridge), so capture the raw text and wrap it (issue #81).
-        # Claude keeps the raw-text stream unchanged.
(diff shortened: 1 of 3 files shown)


alexgreensh:

PR #138, opened 2026-08-15
fix: quote fleet-auditor SKILL.md description (YAML parse failure)

danikdanik:
## What

`fleet-auditor` silently fails to load in strict YAML environments (Devin CLI, `npx skills add`, PyYAML `safe_load`, js-yaml strict mode). The skill is present on disk but never appears in the skills list.

## Why

The `description` field in `skills/fleet-auditor/SKILL.md` contains an unquoted `": "` after `OpenCode)`:

```yaml
description: Cross-system agent token/cost audit (Claude Code, Codex, OpenClaw, Hermes, OpenCode): idle burns, model misrouting, config bloat, with dollar savings.
```

In YAML, `": "` is the mapping separator (YAML 1.2 spec — a colon followed by whitespace in a plain scalar starts a new key). The parser sees `OpenCode): idle burns...` as a nested mapping entry and aborts:

```
yaml.scanner.ScannerError: mapping values are not allowed here
  in "<unicode string>", line 3, column 98:
     ... dex, OpenClaw, Hermes, OpenCode): idle burns, model misrouting,  ...
                                         ^
```

Loaders that fail closed (Devin, `npx skills add`) drop the skill entirely. Claude Code itself fails more quietly — it falls back to treating the file as body content with no metadata, so the skill still works via `/fleet-auditor` but loses its `description` and never auto-triggers. Either way, the author sees no error.

This is a known class of bug across the skills ecosystem, not a Devin-specific quirk:
- [anthropics/skills#359](https://github.com/anthropics/skills/pull/359) — open PR to document the quoting requirement in Anthropic's own skill docs
- [mintlify/docs#6061](https://github.com/mintlify/docs/issues/6061) — same bug breaks `npx skills add`
- [gsd-build/gsd-2#4593](https://github.com/gsd-build/gsd-2/issues/4593) — same bug, two skills dropped silently

## How

Wrap the description value in double quotes:

```yaml
description: "Cross-system agent token/cost audit (Claude Code, Codex, OpenClaw, Hermes, OpenCode): idle burns, model misrouting, config bloat, with dollar savings."
```

No inner double quotes in the string, so no escaping needed. The parsed value is byte-identical to the intended description.

## Files changed

The fix is applied to all three on-disk copies of the file:
- `skills/fleet-auditor/SKILL.md` — canonical source
- `plugins/token-optimizer/skills/fleet-auditor/SKILL.md` — Codex marketplace mirror, generated by `scripts/sync-codex-marketplace-plugin.sh` from `skills/`. Patching it directly here (rather than running the sync script) produces the same result; `mirror-sync` CI passes either way since it regenerates and diffs.
- `cowork/token-optimizer/skills/fleet-auditor/SKILL.md` — Cowork adapter copy, tracked in the repo (hand-maintained alongside the canonical `skills/` copy).

All three are byte-identical before and after the change.

## Verification

- `yaml.safe_load` on the patched frontmatter parses cleanly.
- Scanned every `SKILL.md` in the repo with PyYAML — no other file is affected. The other skills (`token-optimizer`, `token-coach`, `token-dashboard`, `resume-checkpoint`) have no `": "` in their descriptions and already parse fine.
- `CHECKSUMS.sha256` still lists the pre-change hash for this file, but that file is regenerated at release time (last regenerated for v5.11.64) and is already stale from many versions prior to this fix — not a regression introduced here.

## CI

- `mirror-sync`: **pass** — the patched `plugins/` copy matches a fresh regeneration from `skills/`.
- `pytest` (Ubuntu, all versions): **pass**
- `docs`, `installable`, `CLAAssistant`: **pass**
- `pytest-windows` (all versions): **fail** — pre-existing on `main` (same `runtime_env.py` / `claude_home()` / `Path.home()` failure exists on the base branch). Unrelated to this change.

## Suggestion

@alexgreensh, a small unit test in `tests/` that runs `yaml.safe_load` on every `SKILL.md` frontmatter block would permanently prevent unquoted colons from creeping into future skills. Worth considering for a follow-up — keeping this PR trivial to review.


alexgreensh:
Thanks @danikdanik — you found a real bug (the unquoted fleet-auditor SKILL.md description has a mid-value colon that breaks YAML frontmatter parsing). We applied your fix directly to main in 1a7b786 across all three mirrors (root, plugins, cowork), because this PR's branch predates a history rewrite on main and no longer merges cleanly. Credited you in the commit. Closing since the fix has landed.

PR #146, opened 2026-08-19
fix: recognize mise shims in python-launcher.sh

### LINKED ISSUE #145: python-launcher.sh: _is_safe_prefix doesn't recognize mise shim directories — every hook fails with exit 127 for mise users
## Summary

Users who manage Python via [mise](https://mise.jdx.dev) get `token-optimizer: no usable Python 3 interpreter found` on **every** hook, even though a working Python is right there on `$PATH`. Confirmed on Windows (Git Bash), mise 2026.8.8.

## Root cause

`find_interpreter()` walks `$PATH`, finds mise's shim (e.g. `AppData\Local\mise\shims\python3.exe` on Windows), confirms it's executable and non-empty, then gates it through `_is_safe_prefix()`. That allowlist has no case pattern for mise's default shim directory on any platform, so the candidate is rejected. Every `python3`/`python`/`py -3` candidate on `$PATH` gets exhausted the same way, the script falls through to the POSIX-only direct-path fallback list (no Windows entries), and it exits 127.

## Verified on this machine (Windows / Git Bash, mise 2026.8.8)

- `which python3` → `/c/Users/.../AppData/Local/mise/shims/python3.exe` — a real PE32+ executable, not a symlink or batch wrapper, executable and non-empty (passes the `-x`/`-s` checks in `find_interpreter`, fails only `_is_safe_prefix`).
- `mise which python3` → resolves to `AppData/Local/mise/installs/python/3.14.7/python3.exe`, the actual interpreter the shim dispatches to. This path is **not** consulted by the script (it only ever walks `$PATH` directly), so it doesn't need to be allow-listed for this fix.
- No `MISE_DATA_DIR` / `MISE_SHIMS_DIR` override in play — mise is using its default data directory.

## Proposed fix

Add mise's default shim directory as a new case pattern in `_is_safe_prefix()`, hardcoded — not read from `MISE_DATA_DIR`/`MISE_SHIMS_DIR`, to preserve the "prefixes must be hardcoded, never PATH/env-derived" invariant already documented at the top of that allowlist:

- Windows: `/[a-zA-Z]/Users/*/AppData/Local/mise/shims/*`
- macOS: `/Users/*/.local/share/mise/shims/*`
- Linux: `/home/*/.local/share/mise/shims/*`

I have this written and tested on Windows/Git Bash — confirms `python-launcher.sh` now resolves and execs through the mise shim correctly, including the existing `pythonw.exe` flash-avoidance swap (mise ships that twin alongside `python3.exe` too, so it comes along for free). Will open a PR shortly.

## Related, intentionally out of scope for this issue/PR

Other version managers that put a similar shim-style `python3` on `$PATH` would hit this exact same failure mode. Flagging so they're tracked, not bundling them in:

- **pyenv** (macOS/Linux) and **asdf**: shims are plain, extension-less scripts at `~/.pyenv/shims/python3` and `~/.asdf/shims/python3` (or `$ASDF_DATA_DIR/shims`) — same shape as mise's POSIX shims, so the same style of allowlist pattern would likely fix them. Not verified/tested against either, though.
- **pyenv-win**: a bigger gap than the allowlist alone. Its shims are `.bat` files by default, and `find_interpreter()` only tries the `""` and `.exe` extensions — it wouldn't even *discover* a pyenv-win shim regardless of what's allow-listed. Fixing that needs an extension-loop change too, and `exec`-ing a `.bat` from Git Bash has its own argument-quoting quirks worth testing in isolation.

trekie86:
Fixes #145.

## Summary
`_is_safe_prefix()` had no case pattern for [mise](https://mise.jdx.dev)'s default shim directories, so a mise-managed `python3`/`python` on `$PATH` was found by `find_interpreter()` but rejected by the allowlist on every platform.

## Verified
Windows/Git Bash, mise 2026.8.8: `which python3` resolves to `AppData\Local\mise\shims\python3.exe` — a genuine PE executable (not a symlink or batch wrapper), executable and non-empty, passing every check except `_is_safe_prefix`. `mise which python3` resolves to the real interpreter at `AppData\Local\mise\installs\python\3.14.7\python3.exe`, which the script never consults directly (it only walks `$PATH`), so that path doesn't need allow-listing.

Tested against current `main` (5.11.99) after rebasing off it — the newer `TOKEN_OPTIMIZER_PYTHON` override and hook-safety `exit 0` degrade path (970fbc7) are untouched by this diff.

## Change
Adds hardcoded default-shim-directory patterns for Windows, macOS, and Linux, in the same per-user/wildcard-username style already used for `AppData/Local/Programs/Python`. Deliberately **not** read from `MISE_DATA_DIR`/`MISE_SHIMS_DIR`, to preserve the allowlist's documented "hardcoded, never PATH/env-derived" invariant.

## Scope
pyenv/asdf (POSIX, same shim shape) and pyenv-win (`.bat` shims, needs an extension-loop change too) are noted as follow-ups in #145, not included here.

### DIFF
--- a/hooks/python-launcher.sh
+++ b/hooks/python-launcher.sh
@@ -7,6 +7,8 @@
 #   - Windows Store Python (real installs proven alive via a flash-free
 #     GUI-twin proof-of-life probe, console --version probe as the fallback
 #     authority; non-functional AppExecutionAlias stubs skipped automatically)
+#   - mise (https://mise.jdx.dev) shims at their default per-platform data-dir
+#     location (Windows/macOS/Linux)
 # On Windows (Git Bash/MSYS), exec prefers the GUI-subsystem twin over the
 # console binary to avoid the per-hook console-window flash and orphaned
 # conhost.exe: python.exe/python3.exe swap to pythonw.exe, and py.exe (the
@@ -59,6 +61,19 @@ _is_safe_prefix() {
         # Exact filename keeps the anti-hijack intent (no wildcard in that dir).
         /[a-zA-Z]/Windows/py.exe)                                      return 0 ;;
         /[a-zA-Z]/Windows/pyw.exe)                                     return 0 ;;
+        # mise (https://mise.jdx.dev) shim directory, default Windows data dir.
+        # Hardcoded default location only -- never derived from MISE_DATA_DIR/
+        # MISE_SHIMS_DIR, which would reopen the PATH-hijack vector this
+        # allowlist exists to close.
+        /[a-zA-Z]/Users/*/AppData/Local/mise/shims/*)                  return 0 ;;
+    esac
+    # mise (https://mise.jdx.dev) shim directory, default macOS/Linux data dir
+    # (XDG: ~/.local/share/mise/shims). Needs its own case block (not
+    # _SAFE_PREFIXES) because it requires a per-user wildcard component.
+    # Hardcoded default location only, same rationale as the Windows entry above.
+    case "$binpath" in
+        /Users/*/.local/share/mise/shims/*)                            return 0 ;;
+        /home/*/.local/share/mise/shims/*)                             return 0 ;;
     esac
     # C8: case-insensitive WindowsApps allow for Windows-style drive-letter
     # paths. On a case-insensitive FS the dir can be any casing; the drive-


trekie86:

alexgreensh:
Thanks @trekie86 — really precise root-cause writeup, it made this easy to reason about.

Shipped a broader fix in **v5.11.102** (rolled out across v5.11.100–102). Rather than allowlist mise by hand, `_is_safe_prefix` now keeps its hardcoded fast-path and, on POSIX, falls back to an ownership/permission check (interpreter + its symlink-resolved dir owned by you and writable by nobody else — the ssh/sudo/git `safe.directory` boundary). That covers mise, pyenv, asdf, rbenv, and custom `MISE_DATA_DIR`/`PYENV_ROOT` under any home root generically, while a hijacked `python3` in a world-writable or foreign-owned tree stays refused.

Your Windows mise shim pattern is **included and credited** in the drive-letter allowlist (Git-Bash `stat` is unreliable there, so Windows stays hardcoded — exactly as you noted). I also landed the pyenv-win follow-up you flagged: `find_interpreter` now discovers `.bat`/`.cmd` shims on MSYS (the exec path still wants real-Windows confirmation, so if you can validate a pyenv-win `python.bat` actually fires the hook on your box, that'd be gold).

Since the generic gate supersedes the mise-only change, I think this PR can be closed as resolved-by-#145 — but that's your call, I won't close a contributor's PR out from under them. Either way, thank you for the fix and the diagnosis.

trekie86:
Makes sense, that is what I was also concerned about but glad you found a broader solution to accommodate. I'll close the PR.

Issue #139, opened 2026-08-15
UserPromptSubmit: six hook entries spawn 18 processes per prompt (~809ms median blocking); --once-per-session gated after the 1.88MB measure.py import


sadvances:
## Summary

`hooks.json` registers **six** `UserPromptSubmit` entries. Each one runs the full
`python-launcher.sh` → `run.py` → `module_runner.py` → `runpy(measure.py)` chain, so a single
user prompt spawns **18 processes** and imports the 1.88 MB `measure.py` module six times over.

Measured on this machine: **809 ms median / 1000 ms p90 / 1297 ms max** of blocking time on every
prompt. Because `UserPromptSubmit` blocks the request, this is dead air between pressing enter and
the model starting.

Three of the six entries are `--once-per-session` or `--new-session-only`, but that gate is
enforced *inside* `measure.py` (around lines 40543 / 40780) — i.e. **after** the interpreter has
started and the 1.88 MB module has loaded. On every prompt after the first, three of the six
spawns pay the entire startup cost in order to decide to do nothing.

## Environment

| | |
|---|---|
| token-optimizer | 5.11.93 |
| Claude Code | 2.1.233 |
| OS | macOS (Darwin 25.6.0), Apple Silicon |
| Python | 3.14.6 (`/opt/homebrew/bin/python3`) |
| Bytecode cache | warm — `measure.cpython-314.pyc` (1.88 MB) present and current |

## How this was measured

Claude Code records a `durationMs` for each hook invocation in its session transcripts
(`~/.claude/projects/*/*.jsonl`). I aggregated those across the 50 most recent sessions
(~1 day of heavy use, 5 project directories).

I deliberately **did not execute `measure.py`** to time it, since it writes to the plugin's state
directory and would not have been a read-only probe. Every figure below comes either from
harness-recorded durations or from reading the plugin source.

### Blocking hook time, 50 most recent sessions

```
PreToolUse       n=1239 med=  99ms p90= 126ms tot=128.9s :: rtk-rewrite.sh (unrelated, user hook)
UserPromptSubmit n= 146 med= 809ms p90=1000ms tot=119.9s :: python-launcher.sh -> measure.py
Stop             n= 303 med= 244ms p90= 407ms tot= 81.6s :: stop.mjs (unrelated plugin)
SessionStart     n=  61 med= 401ms p90= 527ms tot= 24.6s :: python-launcher.sh -> measure.py
```

[...]

PR #157, opened 2026-08-25
fix(hooks): ensure-health no longer re-adds legacy quality-cache hook (#155)

### LINKED ISSUE #155: ensure-health re-adds legacy quality-cache hook, partly undoing #139
## Problem

On a **script install** (not a plugin install), `measure.py ensure-health` re-installs the pre-5.11.93 legacy `UserPromptSubmit` hook into `~/.claude/settings.json` on every session where it is absent — including sessions where the canonical `hooks/hooks.json` entry is already correctly installed.

The result is that the consolidated dispatcher from #139 and the legacy raw hook both run on every prompt:

```jsonc
"UserPromptSubmit": [
  { "hooks": [{ "type": "command", "command": "for b in bash /bin/bash …; do … exec … hooks/userpromptsubmit_runner.py; done; exit 0" }] },
  { "hooks": [{ "type": "command", "command": "python3 '<install-dir>/skills/token-optimizer/scripts/measure.py' quality-cache --quiet" }] }
]
```

Impact:

- `quality-cache` runs **twice per prompt**, and the second run is an out-of-process re-import of the 1.88 MB `measure.py` — re-introducing a slice of exactly the per-prompt blocking cost #139 removed (18 procs → 3).
- The duplicate entry is **not removable**. Deleting it from `settings.json` works until the next `SessionStart`, when `ensure-health` puts it straight back. I removed it twice and it returned both times.

Found on v5.11.108 (`8cb3a5c`), macOS, script install via `install.sh`.

## Current Behavior

Two predicates decide whether the quality-cache hook is installed, and both test for the literal substring `quality-cache` in the hook's command string:

**`measure.py:41604`** — the `ensure-health` self-heal:

```python
_is_plugin = _is_running_from_plugin_cache() or _is_plugin_installed()
try:
    _qb_disabled = False
    if CONFIG_PATH.exists():
        _qb_cfg = json.loads(CONFIG_PATH.read_text(encoding="utf-8"))
        _qb_disabled = _qb_cfg.get("quality_bar_disabled", False)
    if not _is_plugin and not _qb_disabled and SETTINGS_PATH.exists():
        _sh_settings = json.loads(SETTINGS_PATH.read_text(encoding="utf-8"))
        _sh_hooks = _sh_settings.get("hooks", {}).get("UserPromptSubmit", [])
        if not any("quality-cache" in str(h) for h in _sh_hooks):   # <-- never matches the dispatcher
            setup_quality_bar(quiet=True)
except Exception:
    pass
```

**`measure.py:34283`** — `setup_quality_bar`'s own idempotence check (`_qb_current_state`):

```python
for group in (hooks.get("UserPromptSubmit") or []):
    for hook in (group.get("hooks") or []):
        if "quality-cache" in (hook.get("command") or ""):
            result["hook"] = True
            break
```

Since a299bf7 (`fix(hooks): consolidate 6 UserPromptSubmit entries into one in-process dispatcher (#139)`, v5.11.93), the shipped `hooks/hooks.json` command is:

```
for b in bash /bin/bash /usr/bin/bash /usr/local/bin/bash /opt/homebrew/bin/bash; do
  command -v "$b" >/dev/null 2>&1 && exec "$b" "${CLAUDE_PLUGIN_ROOT}/hooks/python-launcher.sh" \
    "${CLAUDE_PLUGIN_ROOT}/hooks/run.py" hooks/userpromptsubmit_runner.py; done; exit 0
```

The subcommand names moved *inside* `userpromptsubmit_runner.py` (`_sub_quality_cache_warn` at line 206 runs `quality-cache --warn --quiet`), so the string `quality-cache` no longer appears anywhere in the command. Both predicates return `False`, `setup_quality_bar()` sees `current["hook"] == False`, and appends a fresh group at `measure.py:35070`:

```python
hook_cmd = f"python3 '{mp}' quality-cache --quiet"
hook_entry = {"type": "command", "command": hook_cmd}
hook_group = {"hooks": [hook_entry]}
hooks.setdefault("UserPromptSubmit", []).append(hook_group)
```

Flow:

```
SessionStart
   |
   +-- measure.py ensure-health
          |
          +-- probe: any("quality-cache" in str(h)) over settings UserPromptSubmit
          |      canonical entry -> "…/userpromptsubmit_runner.py; done; exit 0"
          |      substring absent -> False  ("hook is missing")
          |
          +-- setup_quality_bar(quiet=True)
                 |
                 +-- _qb_current_state() -> same substring test -> hook: False
                 +-- append  python3 '<mp>' quality-cache --quiet
                                                |
                          every subsequent prompt: dispatcher AND legacy hook both fire
```

Plugin installs escape only by luck: the `not _is_plugin` gate at `measure.py:41595` short-circuits before the probe. The plugin-cache fallback inside `_qb_current_state` (`measure.py:34298`) applies the *same* substring test to the plugin's `hooks.json`, so it would also fail to recognize the shipped canonical hook if it were ever reached.

## Steps to Reproduce

1. Script install v5.11.108, canonical hooks merged into `settings.json` (`python3 skills/token-optimizer/scripts/measure.py setup-all-hooks`).
2. Ensure `UserPromptSubmit` contains exactly one entry — the `userpromptsubmit_runner.py` dispatcher. Remove any legacy `python3 '…/measure.py' quality-cache --quiet` entry.
3. Run the SessionStart hook:
   ```bash
   echo '{"session_id":"t","transcript_path":"","cwd":"'"$PWD"'","hook_event_name":"SessionStart","source":"startup"}' \
     | bash ~/.claude/token-optimizer/hooks/python-launcher.sh \
            ~/.claude/token-optimizer/hooks/run.py \
            skills/token-optimizer/scripts/measure.py ensure-health --once-mark
   ```
4. Re-read `~/.claude/settings.json` → `UserPromptSubmit` now has **2** entries. Repeat step 2–3 and it comes back every time.

## Proposed Solution

Broaden the "is our quality hook installed?" predicate so it recognizes the dispatcher, and share it between the two call sites instead of duplicating the substring test:

```python
_QUALITY_HOOK_MARKERS = ("quality-cache", "userpromptsubmit_runner.py")

def _is_quality_cache_hook(cmd) -> bool:
    cmd = cmd or ""
    return any(m in cmd for m in _QUALITY_HOOK_MARKERS)
```

Use it at `measure.py:34283` and `:34298` (`_qb_current_state`) and at `:41604` (the `ensure-health` self-heal).

Why marker-based rather than subcommand-based: keying installation state on a *subcommand string* breaks every time a subcommand is consolidated behind a dispatcher, which is exactly what #139 did. `_is_our_hook_entry()` (`measure.py:23775`) already solves the equivalent problem for the uninstall sweep by keying on **script basenames** (`our_scripts = ("measure.py", "read_cache.py", "statusline.js", …)`). Reusing that convention here would make the check consolidation-proof rather than needing another marker appended at the next refactor.

A regression test worth adding: assert that `_qb_current_state()` reports `hook: True` when `settings.json` holds the exact `UserPromptSubmit` command from the shipped `hooks/hooks.json`. That single assertion would have caught this at a299bf7.

## Files Involved

- **`skills/token-optimizer/scripts/measure.py`** — `_qb_current_state()` (~L34283, L34298) and the `ensure-health` self-heal block (~L41604): replace the inline `"quality-cache" in …` tests with the shared predicate.
- **`hooks/hooks.json`** — no change needed; it is the correct, canonical definition. (Alternatively, the whole bug closes by appending a `# quality-cache` comment to the command, but that hides the stale predicate rather than fixing it.)

## Additional Context

- **Regression window:** introduced in v5.11.93 / a299bf7 (2026-08-16), the #139 fix. The self-heal predicate itself is older and was not touched by that commit — the command string moved out from under it. Still present in v5.11.108.

- **Workaround** for anyone hitting this before a fix lands — append a no-op comment to the canonical command in `settings.json` so the substring probe is satisfied; the shell ignores it:
  ```
  … hooks/userpromptsubmit_runner.py; done; exit 0  # quality-cache
  ```
  Verified: two consecutive `ensure-health` runs no longer re-add the legacy entry.

- **Separate observation, not part of this report:** `measure.py setup-all-hooks` is additive only. Running it on an installation carrying an older hook set (mine was v5.6.4) reported `Added 18 hook(s), skipped 7` and left 11 superseded entries in place alongside their replacements — `read_cache.py --quiet`, `bash_hook.py --quiet`, `archive_result.py --quiet`, `context_intel.py --quiet` and others each fired twice per event until I removed them by hand. Happy to open that as its own issue if useful.

alexgreensh:
Fixes #155.

## Root cause
Since #139 (a299bf7, v5.11.93) the shipped `UserPromptSubmit` hook is a single in-process dispatcher whose command execs `hooks/userpromptsubmit_runner.py` and runs `quality-cache` **inside** the runner. The literal substring `quality-cache` no longer appears in the hook command.

Three detection sites in `skills/token-optimizer/scripts/measure.py` probed only for that substring, so they treated the consolidated dispatcher as "hook missing":
- UserPromptSubmit self-heal probe (`ensure-health` path)
- SessionStart auto-restore probe (`has_cache_hook`)
- `_is_quality_bar_installed` (settings.json check + plugin-cache `hooks.json` fallback)

`ensure-health` then called `setup_quality_bar`, which appended a fresh legacy `python3 '<mp>' quality-cache --quiet` group. Result: `quality-cache` ran **twice per prompt** (re-introducing a slice of the per-prompt blocking cost #139 removed), and the duplicate was un-removable because the next `SessionStart` re-added it.

## Fix
Add two helpers, `_command_drives_quality_cache(command)` and `_quality_cache_hook_present(groups)`, that recognize **both** the legacy standalone hook and the #139 consolidated dispatcher (marker: `userpromptsubmit_runner.py`). Route all three detection sites through them, so an install whose canonical #139 hook is already present is recognized as "installed" and no legacy hook is appended.

The `setup-quality-bar --uninstall` path is deliberately left matching only `quality-cache`, so it strips the legacy add-on hook without ever removing the legitimate #139 dispatcher (which carries the other UserPromptSubmit subcommands).

## Tests
- New `tests/test_ensure_health_quality_cache_155.py` (6 cases): with the #139 dispatcher already installed, the hook is recognized (`_is_quality_bar_installed` + plugin-cache fallback) and `setup_quality_bar` does **not** re-add a legacy hook. Pins the premise that the dispatcher command has no `quality-cache` substring.
- Full existing suite: **1519 passed, 13 skipped, 0 failed**.
- `plugins/` and `cowork/` mirrors regenerated via the canonical sync scripts; `check-mirror-sync.sh` and `test_cowork_committed_plugin.py` green.

## Note on the cowork README change
`cowork_install.py --emit-committed` (the only supported regen path, required by the anti-drift test) also refreshed `cowork/token-optimizer/README.md`. That mirror was already stale on `main`: #156 updated the root README (BoostGraph scoping) without regenerating the cowork mirror. The refresh is unrelated to the hook logic and rides along because the generator rebuilds the whole tree.

https://claude.ai/code/session_01KmE9PrVMmmCBCbse8Uuap6

### DIFF
--- a/cowork/token-optimizer/README.md
+++ b/cowork/token-optimizer/README.md
@@ -210,8 +210,7 @@ These ten rows are the ones where Token Optimizer is the only 🟢 in the row: w
 
 |  | Token Optimizer | Headroom | RTK | Boost | context-mode | `/context` |
 |---|---|---|---|---|---|---|
-| Compaction survival | 🟢 Progressive checkpoints, restore, tool output digest | 🔴 | 🔴 | — | 🟡 Session guide only | 🔴 |
-| Session continuity | 🟢 Cross-session hints, cold-resume, checkpoint scoring | 🔴 | 🔴 | — | 🟡 Session guide | 🔴 |
+| Session continuity | 🟢 Progressive checkpoints before compaction + restore after, cross-session hints, cold-resume, tool-output digest; measured ~3.9M tokens / ~$19 recovered in a 30-day snapshot (checkpoint restores + lean resumes) | 🔴 | 🔴 | — | 🟡 Session guide only | 🔴 |
 | Structural waste audit | 🟢 Deep per-component (CLAUDE.md, skills, MCP, memory) | 🔴 | 🔴 | 🔴 | 🔴 | 🟡 Summary only |
 | CLAUDE.md and MEMORY.md health | 🟢 8 auditors + attention-curve scoring | 🔴 | 🔴 | 🔴 | 🔴 | 🔴 |
 | Model routing and behavioral coaching | 🟢 12 detectors, subagent cost breakdown, anti-patterns | 🔴 | 🔴 | — | 🔴 | 🟡 Basic suggestions |
@@ -237,11 +236,11 @@ These ten rows are the ones where Token Optimizer is the only 🟢 in the row: w
 | Multi-platform | 🟢 Claude Code, VS Code, Codex, OpenClaw, OpenCode, Hermes, Copilot | 🟢 Claude Code, Cursor, Codex, Aider, Copilot | 🟢 15 integrations | 🟡 Cursor, Claude Code, Copilot, Codex CLI | 🟢 17 integrations | 🔴 Claude Code only |
 | Per-task model and effort advice | 🟢 `route` sizes the task before you spend | — | — | — | — | — |
 | Keep-Warm (cache TTL refresh) | 🟢 Opt-in ping before cache expiry, tripwire auto-off | 🔴 | 🔴 | — | 🔴 | 🔴 |
-| End-to-end task-outcome benchmark | 🔴 Output-token A/B only | — | — | 🟢 Vendor reports Terminal-Bench 2.0 with the same pass rate and ~12% lower cost | — | N/A |
+| End-to-end task-outcome benchmark | 🟡 Controlled A/B on 7 real tasks (output tokens) + measured real-session with/without savings; pass-rate study not yet run | — | — | 🟢 Vendor reports Terminal-Bench 2.0 with the same pass rate and ~12% lower cost | — | N/A |
 | Signed and checksum-verified install | 🟢 `CHECKSUMS.sha256` per release, verified at install, CI-enforced | — | — | 🔴 Installer verifies neither a checksum nor a signature | — | N/A |
 
 </details>
-An em dash means the capability was not verified from first-party material in the 2026-07-26 source audit; it is not a claim that the capability is absent. Boost's "pre-shell" and "shell-level" cells are sourced from its documented wrapper form (`boost <command>`); the mechanism `boost init` uses to wire an agent is not described in its first-party material. Token Optimizer's compression claims are tested against real sessions and an 87-fixture suite you can run yourself. **[Full benchmark methodology and results →](BENCHMARK.md)**
+An em dash means the capability was not verified from first-party material in the 2026-07-26 source audit (Boost's first-party material re-verified 2026-08-26); it is not a claim that the capability is absent. This page scores command-output compression only; JFrog Boost's separately-shipped BoostGraph, a local code-map its agent queries on request, is a different capability and out of scope here. Boost's "pre-shell" and "shell-level" cells are sourced from its documented wrapper form (`boost <command>`); the mechanism `boost init` uses to wire an agent is not described in its first-party material. Token Optimizer's compression claims are tested against real sessions and an 87-fixture suite you can run yourself. **[Full benchmark methodology and results →](BENCHMARK.md)**
 
 <p align="center">
   <a href="https://alexgreensh.github.io/token-optimizer/reference/comparison/"><img src="https://img.shields.io/badge/Read%20the%20official%20full%20comparison-in%20our%20docs-0b7285?style=for-the-badge" alt="Read the official full comparison page in our documentation"></a>
--- a/cowork/token-optimizer/skills/token-optimizer/scripts/measure.py
+++ b/cowork/token-optimizer/skills/token-optimizer/scripts/measure.py
@@ -34260,6 +34260,43 @@ def _is_malformed_to_hook(cmd):
     return has_subshell or has_double_home
 
 
+# Marker substrings that prove a UserPromptSubmit hook already drives the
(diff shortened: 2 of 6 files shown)


alexgreensh:
Landed directly on `main` as `4de9cfd` and shipped in **v5.12.0** (we push fixes straight to main). The change was squashed with a scrubbed message; closing this PR as superseded. Thanks!

PR #158, opened 2026-08-25
fix(dashboard): heal HTML off the hook budget so the daemon-script update always runs

alexgreensh:
## Problem

The dashboard's runway card ("Your plan goes further") kept showing the placeholder

> Your live 5-hour and weekly limit view refreshes while you are actively working — it reappears here on your next session

instead of the real 5h/7d rate-limit savings, and it did **not** self-heal after a plugin upgrade — even though the live meter (`~/.claude/token-optimizer/rate-limits.json`) was healthy and `runway_snapshot()` returned correct windows. This "we shipped the fix but it's still broken" class has recurred several times via symptom patches.

## Root cause (compound strand, verified against current `main`)

`run_ensure_health()` (the SessionStart `ensure-health` hook) runs under an **8s** SIGALRM budget (`_install_hook_budget(8)`). Its two self-heal blocks run in order:

1. **HTML staleness heal**: on a stale sidecar it rebuilt the dashboard **in-process** via `generate_standalone_dashboard(quiet=True, force=True)`. That build is ~9s, which **exceeds the 8s budget**, so it raised `_HookTimeout` — a `BaseException` the local `except Exception` cannot catch.
2. **Daemon-script auto-update**: regenerates `dashboard-server.py` when its baked `TOKEN_OPTIMIZER_DAEMON_VERSION` marker differs from the installed version, then restarts.

Because block 1 blew the budget and `_HookTimeout` propagated, the tick aborted at the outer handler ("hook budget exceeded; skipping ensure-health tick") **before block 2 ran**. On a version bump, **both** the HTML and the daemon script stayed stale. And unlike the CLI `dashboard` path, the in-process call had **no** detached-child fallback to ever catch the HTML up. The runway card renders the placeholder exactly when the injected runway has no live windows (`dashboard.html`: `rwCards ? bars : placeholder`), the state any build captures while the meter is momentarily stale — and a static, never-rebuilt dashboard strands it.

This predates and survives PR #154 (which added a version-downgrade guard; it did not touch the ordering or the budget-bound in-process regen).

## Fix (class-level — make it incapable of staying stranded on a version change)

- **`ensure-health` hands the forced rebuild to the existing DETACHED, unbounded child** (`_spawn_detached_dashboard_selfheal`, now with `force=True`) instead of running it in-process under the budget. The cheap spawn returns at once, so the tick **always reaches the daemon-script auto-update**. The child re-reads a now-healthy meter and fills the windows, clearing the placeholder.
- **Teach the `dashboard` CLI a `--force` flag** and thread it to `generate_standalone_dashboard`, so the detached heal bypasses the 60s write throttle (a stale file a killed regen just wrote must not throttle-skip the heal). The **PR #154 version-downgrade guard still applies** — `force` governs the throttle, not version precedence — so a heal never clobbers a newer dashboard.
- **Make the daemon's synchronous `api/regenerate` robust**: force the dashboard step past the throttle (a human's Regenerate must never silently no-op into serving the same stale file) and **retry once with a doubled cap** on an intermittent lock-contention hang, instead of a single `REGEN_STEP_TIMEOUT` (45s) shot.

`measure.py` is mirrored to all three shipped copies: `plugins/` (direct copy, per `test_duplicate_script_sync`) and `cowork/` via its canonical builder `cowork_install.py --emit-committed` (per `test_cowork_committed_plugin`). The re-emit also refreshed `cowork/.../README.md`, which was already stale on `main` (PR #156 updated the root README but never re-emitted the cowork mirror), incidentally greening a pre-existing red `test_rebuild_emit_committed_leaves_git_clean`.

## Tests

New `tests/test_dashboard_version_selfheal.py` (10 tests): the detached+forced heal argv (and that the default argv is unchanged, preserving the Bug B invariant), the CLI `--force` threading, the `ensure-health` ordering (compound-strand guard — asserts the in-process regen is gone and the daemon update follows the heal), the daemon `api/regenerate` force+retry, and the healthy-vs-stale-meter → windows → placeholder chain. Verified these fail on pre-fix `main`.

**Full suite: `1523 passed, 13 skipped` (`python3 -m pytest tests/`).**

## Security Disclosure

No security-relevant changes. The daemon's localhost/token/origin defenses are untouched; the `api/regenerate` change only adds `--force` and a retry to an already-authenticated handler. The detached heal reuses the existing, security-reviewed `_spawn_detached_dashboard_selfheal` spawn path.

## Agent Disclosure

Claude Code · claude-opus-4-8 did the bulk of the work.

### DIFF
--- a/cowork/token-optimizer/README.md
+++ b/cowork/token-optimizer/README.md
@@ -210,8 +210,7 @@ These ten rows are the ones where Token Optimizer is the only 🟢 in the row: w
 
 |  | Token Optimizer | Headroom | RTK | Boost | context-mode | `/context` |
 |---|---|---|---|---|---|---|
-| Compaction survival | 🟢 Progressive checkpoints, restore, tool output digest | 🔴 | 🔴 | — | 🟡 Session guide only | 🔴 |
-| Session continuity | 🟢 Cross-session hints, cold-resume, checkpoint scoring | 🔴 | 🔴 | — | 🟡 Session guide | 🔴 |
+| Session continuity | 🟢 Progressive checkpoints before compaction + restore after, cross-session hints, cold-resume, tool-output digest; measured ~3.9M tokens / ~$19 recovered in a 30-day snapshot (checkpoint restores + lean resumes) | 🔴 | 🔴 | — | 🟡 Session guide only | 🔴 |
 | Structural waste audit | 🟢 Deep per-component (CLAUDE.md, skills, MCP, memory) | 🔴 | 🔴 | 🔴 | 🔴 | 🟡 Summary only |
 | CLAUDE.md and MEMORY.md health | 🟢 8 auditors + attention-curve scoring | 🔴 | 🔴 | 🔴 | 🔴 | 🔴 |
 | Model routing and behavioral coaching | 🟢 12 detectors, subagent cost breakdown, anti-patterns | 🔴 | 🔴 | — | 🔴 | 🟡 Basic suggestions |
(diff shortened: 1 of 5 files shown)


alexgreensh:
Landed directly on `main` as `cf0cda7` and shipped in **v5.12.0** (we push fixes straight to main). Squashed with a scrubbed message; closing as superseded. Thanks!

PR #166, opened 2026-09-01
Harden hook I/O and runtime resource handling (stdin hang, temp-file planted-file attacks, fd leak, timer restore)

danikdanik:
This groups four small, independent robustness fixes in the hook/runtime layer. They share
the theme of "a hook should never hang, leak, or clobber under adverse conditions," but each
stands alone and can be split into its own PR on request. Each has its own repro below.

## 1. skills/token-optimizer/scripts/hook_io.py - stdin read can block past its own timeout
`read_hook_input` guards the read with `select([sys.stdin], [], [], _STDIN_TIMEOUT)`, then
calls `sys.stdin.buffer.read(max_bytes)`. `select` only guarantees that at least one byte is
available; `read(max_bytes)` then blocks until it gets `max_bytes` or EOF. A host that writes
a partial payload and keeps the pipe open makes the hook hang indefinitely, defeating the
timeout.
Fix: after `select` signals readable, read what is available without blocking (a loop bounded
by the same deadline that still reassembles a chunked payload), rather than a blocking
`read(max_bytes)`.
Repro: pipe a partial JSON payload to the hook and hold the pipe open. The process hangs
today; with the fix it returns after the timeout.

## 2. hooks/module_runner.py - a predictable shared-temp marker is vulnerable to planted-file attacks
`_warn_readonly_scripts_dir_once` writes a once-a-day marker to
`gettempdir()/.token-optimizer-ro-pyc-<sha1(scripts_dir)[:12]>` via `open(marker, "w")`.
The path is world-predictable and lives in the shared temp dir, so on a multi-user host
another local user can pre-plant something hostile there (CWE-377):

- a symlink, and the write follows it to a victim file;
- a FIFO, and `open(..., O_WRONLY)` on a reader-less FIFO blocks the hook indefinitely (the
  function's `except` cannot interrupt a blocked syscall, defeating fail-open);
- a regular file or a hardlink to a victim, and the `"w"` truncation clobbers it.

Fix: remove any existing entry first, then create the marker exclusively, so the code only
ever opens a file it just made:

```python
try:
    os.unlink(marker)          # drops a planted symlink / FIFO / hardlink by name
except OSError:                # (a hardlink's target is left untouched)
    pass
flags = (os.O_WRONLY | os.O_CREAT | os.O_EXCL
         | getattr(os, "O_NOFOLLOW", 0) | getattr(os, "O_NONBLOCK", 0))
fd = os.open(marker, flags, 0o600)
```

`O_EXCL` guarantees we never open a pre-existing entry; `O_NOFOLLOW` and `O_NONBLOCK` close
the tiny unlink-to-open race (a re-planted symlink fails `ELOOP`, a re-planted FIFO fails
`ENXIO` instead of blocking). Every failure here is swallowed by the existing
`except Exception: pass`, so the hook stays fail-open and just skips the warning.
`O_NOFOLLOW`/`O_NONBLOCK` are absent on Windows; the `getattr(..., 0)` fallback makes them
no-ops there. (Moving the marker under a per-user directory the tool owns would also remove
the cross-user race; the exclusive-create approach keeps the existing temp-dir location.)
Repro: on a shared temp dir, pre-plant the marker path as (a) a symlink to a victim file,
(b) a FIFO, or (c) a regular file, then trigger the read-only-scripts-dir branch. Today (a)
follows the symlink, (b) hangs the hook, and (c) truncates the file; after the fix each is
refused and the hook stays fail-open.

## 3. skills/token-optimizer/scripts/measure.py - file descriptor leak on interrupted atomic write
The dashboard-meta atomic write does `tmp_fd, tmp_name = tempfile.mkstemp(...)` then
`os.fchmod(tmp_fd, 0o600)` before `os.fdopen(tmp_fd, ...)`. If `os.fchmod` raises (or a hook
budget `_HookTimeout` fires) between `mkstemp` and `fdopen`, the `except` unlinks the temp
file but never closes `tmp_fd`, leaking the descriptor. Under a repeatedly-timing-out hook
this accumulates.
Fix: close `tmp_fd` on any error before `fdopen` takes ownership of it (once `fdopen`
succeeds, its `with` block owns the close).
Repro: force `os.fchmod` to raise and call the writer in a loop; the process fd count climbs
today and stays flat with the fix.

## 4. hooks/posttooluse_runner.py - restoring an already-expired outer timer disarms it
The nested budget timer saves and restores a previous `setitimer`. On restore it computes
`max(0.0, previous_remaining - elapsed)`. When the outer timer should already have expired
(`previous_remaining - elapsed <= 0`), this passes `0.0` to `setitimer`, which disarms the
timer instead of firing it, so the outer budget is silently dropped.
Fix: detect the `previous_remaining - elapsed <= 0` case explicitly and deliver the outer
budget immediately (via the just-restored handler) rather than round-tripping through
`setitimer` with a value that means "disarm." The delivery is guarded so it never fires into a
default/ignored handler. Arming a tiny positive interval would also fire, but it leaves a
spurious near-zero timer; handling the expired case directly is cleaner.
Repro: nest two budget contexts where the outer is already expired at inner exit. The outer
timeout never fires today; with the fix it fires immediately.

### DIFF
--- a/cowork/token-optimizer/hooks/module_runner.py
+++ b/cowork/token-optimizer/hooks/module_runner.py
@@ -44,7 +44,15 @@ def _warn_readonly_scripts_dir_once(scripts_dir: str) -> None:
             fresh = False
         if fresh:
             return
-        with open(marker, "w", encoding="utf-8") as fh:
+        # Open with O_NOFOLLOW so a symlink pre-created at this predictable,
+        # world-shared temp path by another local user is refused (raises
+        # ELOOP, swallowed by the outer `except Exception` below -> we just skip
+        # the warning) rather than followed to a victim file (CWE-377). O_TRUNC
+        # matches the "w" truncation. O_NOFOLLOW is absent on Windows; the
+        # getattr fallback of 0 makes it a no-op flag there.
+        flags = os.O_WRONLY | os.O_CREAT | os.O_TRUNC | getattr(os, "O_NOFOLLOW", 0)
+        fd = os.open(marker, flags, 0o600)
+        with os.fdopen(fd, "w", encoding="utf-8") as fh:
             fh.write(str(time.time()))
         sys.stderr.write(
             f"[Token Optimizer] note: {scripts_dir} is not writable, so Python "
--- a/cowork/token-optimizer/hooks/module_runner.py
+++ b/cowork/token-optimizer/hooks/module_runner.py
@@ -44,13 +44,23 @@ def _warn_readonly_scripts_dir_once(scripts_dir: str) -> None:
             fresh = False
         if fresh:
             return
-        # Open with O_NOFOLLOW so a symlink pre-created at this predictable,
-        # world-shared temp path by another local user is refused (raises
-        # ELOOP, swallowed by the outer `except Exception` below -> we just skip
-        # the warning) rather than followed to a victim file (CWE-377). O_TRUNC
-        # matches the "w" truncation. O_NOFOLLOW is absent on Windows; the
-        # getattr fallback of 0 makes it a no-op flag there.
-        flags = os.O_WRONLY | os.O_CREAT | os.O_TRUNC | getattr(os, "O_NOFOLLOW", 0)
+        # This marker lives at a predictable, world-shared temp path, so another
+        # local user can pre-plant something hostile there (CWE-377). Defend
+        # against all of it: unlink any existing entry first (drops a planted
+        # symlink, FIFO, or hardlink by name -- a hardlink's target is left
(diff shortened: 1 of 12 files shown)


alexgreensh:
Thanks @danikdanik! All four verified by reproduction before merging: (1) partial payload + held pipe blocked 10s+ against the 0.5s timeout; (2) a planted reader-less FIFO at the marker path hung the hook indefinitely; (3) 50 forced fchmod failures leaked 50 fds; (4) an expired outer budget was silently disarmed for all later handlers. Fixes are minimal and correct, and the Windows no-op fallbacks check out. Regression tests are landing in a follow-up PR. Excellent, careful work.

PR #172, opened 2026-09-02
feat(core): optional local post-flush extension loader

alexgreensh:
## Summary
- The session-end flush worker can now run one optional local extension: `<config dir>/extensions/post_flush.py` defining `run(context)`.
- OFF by default (no file, no load, no I/O beyond one existence check); loaded ONLY from that exact config-dir path (no env override, no repo/snapshot paths); group/world-writable files are ignored (local code injection guard).
- Fail-open end to end: extension exceptions, broken modules, missing `run` — all swallowed; the hook always completes and the flush lock is always released.
- Budgeted: the extension receives `context["time_left_fn"]` (seconds left in the flush budget) and is documented to defer work that does not fit; the existing 20s watchdog still governs.
- Neutral docs: `docs/local-extensions.md` + one sentence in PRIVACY.md. Token Optimizer ships no extensions.

## Test plan
- [x] tests/test_post_flush_extensions.py (9): off by default (no open() calls), config-dir-only (decoy + env override ignored), context keys, fail-open on exception/broken module/missing run, world-writable ignored, worker completes + releases lock, worker runs extension after flush
- [x] Mirror parity tests green after regeneration (67 passed across the parity suites)
- [x] scripts/check_docs_claims.py + check_readme_integrity.py green
- [ ] Full suite re-run at merge time (Track A sequences the merge)

### DIFF
--- a/PRIVACY.md
+++ b/PRIVACY.md
@@ -118,6 +118,8 @@ Manual deletion: remove the following directories:
 
 ## Cross-Platform
 
+Locally installed extensions (see `docs/local-extensions.md`) may run inside the session-end flush worker; Token Optimizer ships none by default.
+
 The same privacy guarantees apply across all supported platforms (Claude Code, Codex, OpenCode, Hermes). Data paths vary by platform but the architecture is identical: local-only, zero network, credential-redacted storage.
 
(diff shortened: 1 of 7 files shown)


alexgreensh:

PR #173, opened 2026-09-02
test: widen lease-expiry race window in owner-hard-exit parity test

alexgreensh:
## Summary
- test_owner_hard_exit_recovers_only_after_lease_expiry raced: the child's interpreter startup consumed most of the 0.25s lease before the parent's 'immediate' acquire, so on slow runners (observed: Windows CI, run 33594427987, and once in a local full-suite run) the lease was already expired and immediate came back True.
- Fix: lease_seconds 0.25 -> 2.0, sleep 0.3 -> 2.2. Both assertions become deterministic; the child still hard-exits holding the lease, so the guarded invariant is unchanged.
- Test-only; product untouched.

## Test plan
- [x] Test passes locally (2.57s)
- [ ] CI all legs green

### DIFF
--- a/tests/test_defeat_exit_cleanup.py
+++ b/tests/test_defeat_exit_cleanup.py
@@ -812,6 +812,13 @@ def test_defeat_bug_c_reclaim_on_already_unlinked_lease_no_raise(tmp_path):
     )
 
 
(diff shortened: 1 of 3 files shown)


alexgreensh:

PR #174, opened 2026-09-02
test: regression tests for #166 hook robustness fixes

alexgreensh:
## Summary
Four regression tests for the hook-robustness fixes merged in #166 (686847b), each reproducing a bug demonstrated live on main before the merge (full evidence: sessions/2026-09-01-parallel-sprint/reports/A-PR-166-verdict.md):
1. stdin partial payload + held pipe must return within the deadline (hung 10s+ on main)
2. planted reader-less FIFO at the ro-pyc marker path must fail open, not hang (hung indefinitely on main)
3. dashboard-meta write must not leak tmp_fd when fchmod raises (50 fds leaked in 50 calls on main)
4. an expired outer budget must be delivered via the restored handler, not disarmed by setitimer(0.0)

All four fail on pre-#166 main and pass on current main. POSIX-only tests are win32-skipped with reasons (mkfifo/setitimer/fchmod availability).

## Test plan
- [x] All 4 pass locally on current main (2.56s)
- [ ] CI all legs green

### DIFF
--- a/tests/test_hook_robustness_regressions.py
+++ b/tests/test_hook_robustness_regressions.py
@@ -1,0 +1,200 @@
+#!/usr/bin/env python3
+"""Regression tests for PR #166 (hook I/O and runtime resource hardening).
+
+Each test reproduces a bug that was demonstrated live on main before #166
+landed (evidence in the PR description and sessions/2026-09-01-parallel-sprint/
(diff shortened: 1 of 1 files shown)


alexgreensh:

Issue #180, opened 2026-09-11
Windows Codex hooks fail with “hook exited with code 1” during runtime resolution


Exclusive2800:
## Summary

  On native Windows, `codex_install.py` can generate a broken Codex hook command for versioned marketplace installs.

  The command assigns `TOKEN_OPTIMIZER_RUNTIME_ROOT` inside a `for /f` loop, then references it with delayed expansion:

>   ```cmd
>   for /f ... %R ... do @set "TOKEN_OPTIMIZER_RUNTIME_ROOT=<base>\%R" && python.exe "!TOKEN_OPTIMIZER_RUNTIME_ROOT!\hooks\run.py"

cmd.exe expands !TOKEN_OPTIMIZER_RUNTIME_ROOT! before the loop assignment, so Python receives the literal placeholder path.

  Codex reports:

>   Stop hook (failed)
>   error: hook exited with code 1

  ## Environment

  - Windows native command hooks
  - Codex CLI 0.154.0
  - Token Optimizer 5.13.11
  - Versioned marketplace cache layout: token-optimizer\<version>\

  The issue was first observed with Token Optimizer 5.13.2. The 5.13.11 generator still emits the same problematic command shape.

  ## Steps to reproduce

  1. Install Token Optimizer through the Codex marketplace on Windows.
  2. Run codex-install with a versioned marketplace cache.
  3. End a Codex session to trigger the Stop hook.
  4. Observe the generic hook failure message.
  5. Inspect the generated command in %USERPROFILE%\.codex\hooks.json.
  6. Remove the final >NUL 2>&1 redirection and execute the command through cmd.exe /d /c with {} on standard input.

  Observed error:

>   python.exe: can't open file '...\!TOKEN_OPTIMIZER_RUNTIME_ROOT!\hooks\run.py':
>   [Errno 2] No such file or directory

  The direct command exits with code 2; Codex surfaces it as hook exit code 1.

  ## Expected behavior

  The generated command should resolve the active version directory and invoke:

>   <resolved-runtime>\hooks\run.py

  The hook should exit successfully and perform the documented Stop-hook checkpoint and session flush.

  ## Actual behavior and impact

  The runtime placeholder is passed literally to Python. The hook runner never starts, so session checkpointing, metrics flushing, and related Token Optimizer updates are skipped. Because stderr is redirected to NUL, users only see the unhelpful Codex message.

  The affected implementation is _hook_command in:

[...]

PR #184, opened 2026-09-15
fix(codex): replace base64 Windows hook bootstrap with AV-safe launcher file (#183)

### LINKED ISSUE #183: codex_install.py: base64 exec bootstrap trips antivirus signatures (flagged as Metasploit variant)
The Windows launcher added in c26296033 generates a hook command of this form:

    python -c "exec(__import__('base64').b64decode('<payload>'))"

SentinelOne flags `skills/token-optimizer/scripts/codex_install.py` from 5.13.14 (sha256 `e2086721cc853591816a85e11a09fea3d0d9cc0a1fb30ad51364a21dcb705c0c`) as a Metasploit variant, once per copy of the file shipped in the package (3 detections). `exec` + `b64decode` is a generic signature for obfuscated Python loaders, so the match is on the literal string rather than on behavior.

For the record, the file itself looks clean: no network primitives anywhere, and the only `*eval` is `ast.literal_eval`. On macOS the launcher is never even reached, since it is gated on `sys.platform == "win32"` - the scanner is reading the source statically, not observing execution.

Still worth avoiding, because a signature hit on an install script surfaces as a malware detection rather than a warning, and that is expensive for anyone running EDR.

Suggestion, if it fits the constraint from #180: write the bootstrap to a real `.py` file at a stable path at install time and invoke it by path (`python "<stable>\launcher.py" <script> <args>`) instead of embedding it in the command line. Paths and arguments then live inside the file or as plainly quotable argv, so the cmd.exe single-parse problem stays solved and no encoded-exec string appears in the command at all. The bootstrap is already deterministic and auditable through `--decode-launcher`, so writing it out loses nothing.

alexgr-agent:
Fixes #183.

## What and why

On versioned marketplace installs, `codex_install.py` baked Windows hook commands as `python -c "exec(__import__('base64').b64decode(...))"`. The payload was fixed, readable, and decode-auditable, but the literal encoded-exec string trips generic-loader antivirus signatures (reported: SentinelOne flagging the installer as a Metasploit variant, once per shipped copy of the file). A signature hit on an install script surfaces as a malware detection, which is expensive for anyone running EDR.

The encoded bootstrap is replaced by a plain, auditable launcher file. The installer copies `hooks/windows-launcher.py` next to the versioned install dirs (`.../token-optimizer/windows-launcher.py`, a path that survives the marketplace swapping the version directory on upgrade) and bakes a command that invokes it by quoted path with plainly quoted argv. No encoded-exec string appears anywhere, and no cmd.exe expansion form is needed, so the single-parse constraint from #180 stays solved.

## Behavior retained

- **Version resolution:** newest semver sibling of the install parent, fail-open to `--baked-root`, `TOKEN_OPTIMIZER_DEBUG`-gated resolver log on fallback.
- **stdin/argv/env passthrough:** the runner sees `sys.argv = [run.py, script, *args]`, per-hook env applied, stdin untouched; the trailing ownership marker is stripped before dispatch.
- **Legacy-command recognition:** base64-era and for/f-era commands are still claimed by reinstall/uninstall ownership checks; `--decode-launcher` still decodes base64-era commands, and points launcher-file commands at their plain source.
- **Upgrade trust semantics:** `_launcher_signature` now normalizes only the `--baked-root` version leaf; every other part of the command compares verbatim, so tampering still forces review. Legacy base64 commands compare verbatim against the new shape, so the next install replaces them (the one-time review that ships this fix).

The launcher copy is atomic (unique same-directory temp file + replace, always cleaned up), idempotent, written before `hooks.json` so a failed write aborts the install loudly instead of leaving hooks pointing at a missing file.

## Compatibility notes

- **Percent expansion limitation (new relative to the base64 form):** cmd.exe expands `%...%` pairs across the entire `/C` command line, even inside quotes. A percent-bearing install path (legal but rare) can be altered before the launcher sees it; the retired base64 form did not have this limitation because paths lived inside the encoded payload. Mitigation: use an install path without `%`.
- **Uninstall intentionally leaves `windows-launcher.py` in place**: it is shared across global and project installs and is harmless plain source.
- **Windows cmd.exe execution is covered by the Windows CI leg** (the generated command runs through a real `%COMSPEC% /C` with metachar-laden paths, argv, env, and stdin, across an upgrade). It cannot be exercised on non-Windows hosts; the cross-platform tests execute the launcher directly on every OS.
- Generated arguments and env values cannot contain embedded double quotes (Windows forbids `"` in file names and the generated values are fixed), so no general embedded-quote handling is claimed.

## Also fixed in passing

The command-compression shell resolver crashed with `PermissionError` on unreadable PATH entries (e.g. another user's `~/.cargo/bin` on a shared machine); unreadable entries are now skipped like any other non-match, with a regression test.

## Tests

- New: AV-safe source-string guards (the generator and the launcher neither contain nor emit `exec(`/base64/bootstrap patterns); launcher install atomicity, idempotence, temp-uniqueness, and failure-path cleanup tests; install()-level ordering and dry-run tests; cross-platform launcher execution tests (metachar argv/env/stdin, marker stripping, semver sort, fallback log); legacy decode/recognition/signature tests; cmd.exe execution tests on the Windows CI leg.
- Full suite on this change vs pristine main in the same environment: **zero new failures** (3548 passed / 13 failed vs 3479 passed / 78 failed on pristine main; the fixed delta is the 66 environment-caused `PermissionError` failures resolved above, and the remaining failures reproduce on pristine main: `tomllib` on Python 3.10, an interpreter trust gate, and one intermittently flaky savings-collector test that also fails standalone on unmodified main).
- `plugins/` and `cowork/` mirrors regenerated and byte-identical (mirror-sync gate passes).

Version bump and release notes are left for the release commit; the CHANGELOG entry sits under `[Unreleased]`.


alexgr-agent:
Superseded by #186, which ships this fix plus follow-up hardening and is now merged into main. Closing unmerged.

PR #185, opened 2026-09-15
fix(hermes): context-fill nudge must use the live prompt, not the cumulative tally

NoTimeforInfinity:
## Problem

The Hermes context-fill nudge measures the **session-cumulative** input tally instead of the live
prompt, so it eventually reports a context emergency that does not exist.

`hermes/__init__.py` accumulates every call's `input_tokens` into `_TALLY[session]["input"]`
(`on_post_api_request`), and the nudge divides that sum by the model window
(`on_pre_llm_call` → `current_input = tally["input"]`). Hermes re-sends the whole conversation on
every turn, so that sum climbs past the window regardless of how small the live context is.

## Evidence (Hermes v0.21.x, session with 129 API calls)

| figure | value |
|---|---|
| plugin's cumulative tally | **1,285,803** tokens |
| what the nudge printed | `Context ~100% full (~1,285,803 input tokens vs assumed 1,000,000 window) Grade: F` (percentage capped at 100) |
| what Hermes itself reported for the same session | **278,545 / 1,000,000 = 28%** |

The host's own number was right; the nudge was ~4.6x over, and the `F` grade is computed from the
same figure, so the grade was wrong too. (The window itself resolved correctly — the model window
was read from Hermes's config, so only the numerator was wrong.)

## Fix

* `on_post_api_request` also records `last_prompt = input_tokens + cache_read_tokens` for the call
  it just saw. Cached prompt tokens are still prompt tokens — they occupy the same window — so
  excluding them would undersell the fill by roughly the cache-hit ratio.
* The nudge uses `last_prompt`, falling back to the existing history estimate when no call has
  completed yet (first turn).
* Message wording changed to match what is actually measured:
  `(last request prompt ~N tokens vs model window M)`.
* The cumulative tallies are untouched for cost/usage reporting and the dashboard.

## Tests

`tests/test_hermes_context_fill_nudge.py` (new, 5 cases). Against the previous code **3 of the 5
fail**, against this branch all 5 pass:

```
$ python -m pytest tests/test_hermes_context_fill_nudge.py -q      # before
3 failed, 2 passed

$ python -m pytest tests/test_hermes_context_fill_nudge.py -q      # after
5 passed
```

Cases: 200 small turns whose cumulative tally passes the window but whose live prompt is tiny (must
stay silent); a genuinely full live prompt (must nudge and report the live number, never the
cumulative one); cached-only prompt growth (must count toward fill); once-per-session gating; the
first-turn history-estimate fallback.

Wider check: `tests/test_counted_cumulative.py tests/test_hermes_fleet.py
tests/test_bridges_no_window.py` + the new file → **149 passed, 1 skipped**. No other test in the
suite references `_TALLY` or `on_pre_llm_call`, so nothing else depended on the old semantics.

## Notes

Applied and verified on a live Hermes install first (the local plugin payload was patched ahead of
this PR, with the original file backed up), which is where the measurements above come from. Docs:
none needed — no user-facing configuration changed.


alexgreensh:
Thanks for this fix, @NoTimeforInfinity, it's reviewed, green, and ready. We've staged it to ship (with a small follow-up that adds cache-write tokens to the live figure so a cache-creation turn isn't undercounted). The only thing holding it is our CLA: could you accept it via the CLAAssistant comment above? Once it's signed we'll include it in the next release. Much appreciated.

alexgreensh:
Thanks @NoTimeforInfinity — your fix is correct and ships. Closing this in favor of #187, which carries your change rebased onto `main` (the only conflict was CHANGELOG) so full CI runs, plus a reviewed follow-up: the live figure must sum `input + cache_read + cache_write` (dropping `cache_write` undercounts a cache-creation turn), and a guard so an empty-usage response can't reset the reading to 0. Credit is preserved in the commit history on #187.

PR #187, opened 2026-09-16
fix(hermes): context-fill nudge uses live prompt (input+cache_read+cache_write), not cumulative tally

alexgreensh:
Ships PR #185 by @NoTimeforInfinity from an origin branch so full CI runs, plus one follow-up found while verifying:

- Original fix: the Hermes context-fill nudge divided the session-CUMULATIVE input tally by the model window, so it eventually reported a false ~100%/Grade-F context emergency. Now uses the live last-request prompt.
- **Follow-up:** the live figure summed only `input + cache_read`, dropping `cache_write`; every prompt token occupies the window and Hermes' CanonicalUsage.prompt_tokens = input + cache_read + cache_write, so a cache-CREATION turn read too low. Sum all three; adds a cache-write regression test.

Supersedes #185. All CI green (the one earlier red was a confirmed pre-existing Windows lease-lock flake, unrelated; passed on re-run).

Co-authored-by: NoTimeforInfinity


alexgreensh:
Updated: rebased onto `main` (only CHANGELOG conflicted, both sides kept). Core live-prompt logic verified against upstream `CanonicalUsage.prompt_tokens` (input + cache_read + cache_write), so there is no double-count. Refinements applied on review: (1) an empty-usage response no longer resets `last_prompt` to 0 and silences a real ~full-window warning — it prefers the host's canonical `prompt_tokens` with a component-sum fallback and only overwrites on a positive prompt; (2) removed dead `tally['calls']` state; (3) tightened two loose test assertions and added regression tests. Full suite green; mirror-sync green.

Issue #189, opened 2026-09-16
Bash cross-turn dedup is keyed by session_id, so a subagent gets "identical to your previous output" for output it never saw


kaminskypavel:
## Summary

`bash_compress_hook.py` collapses a repeated Bash command into
`[Token Optimizer: identical to your previous `<cmd>` output this session; unchanged.]`
plus an `expand` pointer. The lookup is keyed by the hook payload's `session_id`.
In Claude Code, subagents share their parent's `session_id`, so when the
orchestrator or a sibling subagent already ran the same command, a subagent
receives the "identical" note for output that was never in its own context. It
has to spend a turn on `measure.py expand` to get the content, or it carries on
without it.

## Environment

- token-optimizer 5.13.14 when observed; the same code path is on `main` at
  5.13.15 (`689cba8`)
- Claude Code 2.1.273, macOS
- Workflow: one orchestrator session fanning out parallel background subagents
  with the Agent tool

## Where

- `skills/token-optimizer/scripts/bash_compress_hook.py:77`: the payload
  `session_id` is exported as `CLAUDE_SESSION_ID`.
- `bash_compress_hook.py:440` and `:466`: `_crossturn_dedup` opens
  `SessionStore(session_id)` and looks up the prior run by command hash only.
- `bash_compress_hook.py:473`: the recency guard assumes a run from the last
  hour is "probably still in the agent's context". That does not hold when the
  prior run belongs to a different agent.
- For contrast, `read_cache.py:1127` already scopes by
  `hook_input.get("agent_id") or hook_input.get("session_id")`.

Claude Code populates `agent_id` and `agent_type` in `PostToolUse` input when the
hook fires inside a subagent (hooks reference, "Inputs").

## Observed

Across two orchestrated sessions (about 125 subagents), I matched every
"identical to your previous" tool result against the transcripts:

- 31 cases (16 and 15) where the only earlier run of that exact command was by a
  different agent in the same session.
- Example: two doc-review subagents started in parallel. One ran
  `cat <shared brief>.md`; 9 seconds later its sibling ran the same command and
  got the "identical" note instead of the file.
- In the first session, 10 of those notes were followed by an `expand` call.

The time cost is small (one extra turn each). The larger problem is
correctness: a subagent that does not expand works without the content it asked
for.

## Repro (from the code path)

1. In a Claude Code session with token-optimizer enabled, run a read-only Bash
   command whose output is at least 200 characters, for example `cat README.md`.
[...]

Issue #188, opened 2026-09-16
Windows launcher: guard runner path before runpy (POSIX parity)


alexgreensh:
Found via torture-room while shipping 5.13.15 (PR #186).

**What**: `hooks/windows-launcher.py` resolves the newest semver sibling of the install parent, then `runpy.run_path`s `<resolved>/hooks/run.py` without checking the runner exists or is readable. If the newest sibling is an incomplete version dir (an upgrade mid-extraction, or a partially-pruned dir), the hook raises `FileNotFoundError` and 127s instead of failing open.

**POSIX already guards this**: the bash resolver does `T="<root>hooks/run.py"; [ -r "$T" ] || exit 0` before exec. The Windows launcher has no equivalent.

**Severity**: low. Pre-existing parity gap (the retired base64 launcher behaved the same), so it is not a 5.13.15 regression. Real trigger is an upgrade race.

**Fix**: when resolving the newest semver sibling, skip any dir lacking a readable `hooks/run.py`; fall open to the next-newest, then to `--baked-root`. Mirror the POSIX `[ -r ]` intent. Add a cross-platform test where a newer-but-incomplete sibling must be skipped.

Proposed for 5.13.16 (alongside the hermes context-fill fix, PR #187).

alexgreensh:
Fixed by #191 (merged): the Windows hook launcher now admits only semver runtime dirs whose run.py actually opens (newest-first, skips incomplete/unreadable/zero-byte), falling back to the baked root. Ships in the next release.

PR #190, opened 2026-09-17
Fix cross-agent collisions in Bash cross-turn dedup (#189)

alexgr-agent:
# Fix cross-agent collisions in Bash cross-turn dedup (#189)

## What changed

Cross-turn dedup now scopes its `SessionStore` to the model-visible agent context. The current PostToolUse payload always overrides any inherited dedup-agent environment value for the duration of the call, then restores the exact previous state on success or exception. Explicit agent ids use a session-plus-hashed-agent store identity.

Missing, `None`, or empty `agent_id` retains the historical `SessionStore(session_id)` identity exactly. Whitespace is treated as a supplied opaque agent id rather than normalized to missing.

Only dedup state is agent-scoped. Archives and savings accounting remain session-scoped. Command hashing, raw-byte identical detection, reference wording, credential redaction, retention, and fail-open behavior are unchanged.

## Tests

- New sibling-agent and same-agent isolation tests.
- Main sentinel and legacy store-identity coverage.
- Current payload overrides stale inherited context.
- Success and exception restoration with initially absent and initially present environment state.
- Whitespace opaque-id behavior.
- Final combined focused run after the judge fix: 51 passed, 2 skipped.
- Judge-reproduced evidence from the prior package: focused 47 passed, 2 skipped; broad 530-test shard identical on pristine and changed trees.

## Local coverage limits

Windows, macOS, and Windows ACL behavior were not run locally. The launcher-specific disclosure belongs to #188; this PR's hook tests ran on Linux.


alexgr-agent:

alexgreensh:
On review, a gap was found beyond the dedup path: the same hook's `thrash_guard.check` still keyed on the bare `session_id`, so a subagent could receive a "has run N times this session" nudge for commands only the main agent ran. Fixed: `thrash_guard.check` now takes an agent-scoped store id (default preserves prior behavior), wired from the same identity the dedup path uses. A supplied `agent_id` is treated as an opaque identity (never normalized or merged); a missing, empty, or whitespace-only id falls back to the historical session-only identity. Edit-detection reads the shared session activity log, so a subagent that edits a file between identical runs still suppresses the false "stuck in a loop" nudge. Full suite green; mirror-sync green.

alexgreensh:
**Disclosure correction (supersedes the whitespace note in the PR body).** A review caught that the first hardening commit over-reached: it `.strip()`-normalized *every* agent_id and hashed the stripped value, so ` agent-A ` and `agent-A` collided into one store — contradicting the opaque-identity contract. Corrected in `8e80b10c`: strip is now consulted **only** to decide emptiness. Final contract: whitespace-only / None / empty agent_id falls back to the main (bare session) identity (so a host sending ` ` can't silently fork main off its dedup history); any non-empty agent_id is treated as an **opaque** identity, hashed un-normalized, so ` agent-A`, `agent-A`, and `agent-A ` are three distinct stores. Added collision tests proving they don't merge. 112 targeted passed; mirror-sync green.

PR #191, opened 2026-09-17
Skip incomplete Windows runtime versions in the hook launcher (#188)

alexgr-agent:
# Skip incomplete Windows runtime versions in the hook launcher (#188)

## What changed

The Windows version resolver now admits only semver directories whose `hooks/run.py` is a readable regular file. An incomplete or unreadable newest upgrade is skipped for the next-newest healthy version, then existing baked-root fallback behavior. Argv, stdin, environment forwarding, marker parsing, and debug-only fallback logging are unchanged.

## Tests

- Incomplete newest upgrade falls through to the next healthy version.
- Unreadable POSIX mode-bit proxy falls through where supported.
- Existing newest-version, baked fallback, argv/stdin/env, and launcher security tests remain green.
- Final combined focused run after the judge fix: 51 passed, 2 skipped.
- Judge-reproduced evidence from the prior package: focused 47 passed, 2 skipped; broad 530-test shard identical on pristine and changed trees.

## Local coverage limits

Windows, macOS, and Windows ACL behavior were not run locally. Linux subprocess tests are a behavioral proxy, Windows CI is decisive, and unreadable POSIX mode bits do not establish Windows ACL coverage.


alexgr-agent:

alexgreensh:
On review, two defects were found and fixed in the version-admission gate: (1) the readability check was a no-op on Windows — it checked the read-only attribute, not ACLs, on the one platform this launcher runs on; (2) the scan could abort on the first unreadable candidate, because `Path.is_file()` raised `PermissionError` that escaped into the outer `except OSError` on some Python versions, selecting the wrong runtime. Both are fixed by sorting candidates newest-first and picking the first whose `run.py` actually opens (per-candidate try/except), which also skips zero-byte runners. Argv, stdin, and environment forwarding are unchanged. Full suite green; mirror-sync green.

